PR #9, opened 2016-01-21
Add sqlite backend option

phyrog:
This PR adds the possibility to use SQLite3 as a backend for postfixadmin. Since Postfix and Dovecot both already support it, postfixadmin is the last piece of the SQLite puzzle.

I had to introduce some special cases for sqlite here and there, but mostly it's just another case, where there is already a distinction from mysql and postgres.

Other times I had to work around missing features in sqlite (especially the `..._num_rows` function).

Each translation file now also includes a `dateformat_sqlite` entry.

One of the bigger changes is in `upgrade.php`, where I renamed all the upgrade functions that were used for both mysql and postgres with a suffix `_mysql_pgsql`, so that they are only run for mysql and postgres, but not for sqlite. I also introduced the current database schema in only one function for sqlite, since there is no need for older database schemas for sqlite.

The only thing that does not currently match its mysql counter part is the `cache` column in the `vacation` table, as I added a default value (there is a bug report on this on the tracker for mysql as well).

So, if I missed anything or violated any style guidelines, please let me know and I'll fix it.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -80,6 +80,7 @@ function language_hook($PALANG, $language) {
 // mysql = MySQL 3.23 and 4.0, 4.1 or 5
 // mysqli = MySQL 4.1+ or MariaDB
 // pgsql = PostgreSQL
+// sqlite = SQLite 3
 $CONF['database_type'] = 'mysqli';
 $CONF['database_host'] = 'localhost';
 $CONF['database_user'] = 'postfix';
@@ -90,6 +91,8 @@ $CONF['database_name'] = 'postfix';
 // If you need to specify a different port for POSTGRESQL database connection
 //   uncomment and change the following
 // $CONF['database_port'] = '5432';
+// If sqlite is used, specify the database file path:
+//   $CONF['database_name'] = '/etc/postfix/sqlite/postfixadmin.db'
 
 // Here, if you need, you can customize table names.
 $CONF['database_prefix'] = '';
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -302,6 +302,9 @@ function escape_string ($string) {
         if ($CONF['database_type'] == "mysqli") {
             $escaped_string = mysqli_real_escape_string($link, $string);
         }
+        if (db_sqlite()) {
+            $escaped_string = SQLite3::escapeString($string);
+        }
         if (db_pgsql()) {
             // php 5.2+ allows for $link to be specified.
             if (version_compare(phpversion(), "5.2.0", ">=")) {
@@ -490,7 +493,9 @@ function create_page_browser($idxfield, $querypart) {
     if (db_pgsql()) {
         $initcount = "CREATE TEMPORARY SEQUENCE rowcount MINVALUE 0";
     }
-    $result = db_query($initcount);
+    if (!db_sqlite()) {
+        $result = db_query($initcount);
+    }
 
     # get labels for relevant rows (first and last of each page)
     $page_size_zerobase = $page_size - 1;
@@ -507,6 +512,15 @@ function create_page_browser($idxfield, $querypart) {
             ) idx WHERE MOD(idx.row, $page_size) IN (0,$page_size_zerobase) OR idx.row = $count_results
         "; 
     }
+
+    if (db_sqlite()) {
+        $query = "
+            WITH idx AS (SELECT * $querypart)
+                SELECT $idxfield AS label, (SELECT (COUNT(*) - 1) FROM idx t1 WHERE t1.$idxfield <= t2.$idxfield) AS row
+                FROM idx t2
+                WHERE (row % $page_size) IN (0,$page_size_zerobase) OR row = $count_results";
+    }
+
 #   echo "<p>$query";
 
 # TODO: $query is MySQL-specific
@@ -1249,6 +1263,17 @@ function db_connect ($ignore_errors = 0) {
         } else {
             $error_text .= "<p />DEBUG INFORMATION:<br />MySQL 4.1 functions not available! (php5-mysqli installed?)<br />database_type = 'mysqli' in config.inc.php, are you using a different database? $DEBUG_TEXT";
         }
+    } elseif (db_sqlite()) {
+        if (class_exists ("SQLite3")) {
+            if ($CONF['database_name'] == '' || !is_dir(dirname($CONF['database_name'])) || !is_writable(dirname($CONF['database_name']))) {
+                $error_text .= ("<p />DEBUG INFORMATION<br />Connect: given database path does not exist, is not writable, or \$CONF['database_name'] is empty.");
+            } else {
+                $link = new SQLite3($CONF['database_name']) or $error_text .= ("<p />DEBUG INFORMATION<br />Connect: failed to connect to database. $DEBUG_TEXT");
+                $link->createFunction('base64_decode', 'base64_decode');
+            }
+        } else {
+            $error_text .= "<p />DEBUG INFORMATION:<br />SQLite functions not available! (php5-sqlite installed?)<br />database_type = 'sqlite' in config.inc.php, are you using a different database? $DEBUG_TEXT";
+        }
     } elseif (db_pgsql()) {
         if (function_exists ("pg_pconnect")) {
 			if(!isset($CONF['database_port'])) {
@@ -1299,7 +1324,7 @@ function db_get_boolean($bool) {
             return 't';
         }  
         return 'f';
-    } elseif(Config::Read('database_type') == 'mysql' || Config::Read('database_type') == 'mysqli') {
+    } elseif(Config::Read('database_type') == 'mysql' || Config::Read('database_type') == 'mysqli' || db_sqlite()) {
         if($bool) {
             return 1;  
         } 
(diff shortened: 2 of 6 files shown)


palepurple:
Hi - thanks for submitting this. 

I'll try and find time to test it locally before I (?) merge it. 

At a quick glance everything looks good.

cboltz:
Thanks for the patch. It's good to see that adding sqlite support worked without adding too many special cases :-)

The patch looks good, with some small exceptions:

a) some notes about setup.php:
- traditionally, the function numbers correspond to the SVN revision on sourceforge, but we might drop that tradition with the changed upgrade handling. If you want to follow it, use 1824 for creating the sqlite tables.
- a revision number is for one change (if needed, split by database type), for example "add field foo to table bar"
- that said - upgrade_1795_sqlite() collides with the unrelated upgrade_1795_mysql(), so please rename upgrade_1795_sqlite() to a higher number
- the fast-forward to r1794 [number needs to be adjusted, see previous note] makes sense, but 
    - you should add a comment explaining that ("sqlite support was introduced in r[the number you choose], so there's no need to scan for older upgrade functions") and
    - you should do that before entering the loop, which probably means to modify $current_version instead of $i
- the added `error_log("checking " . $i);` looks superfluous - forgotten debug code?
- there are some superfluous spaces in $replace for sqlite at the end of some lines. You probably copied them from the mysql or pgsql $replace, so it's not really your fault - nevertheless I'd prefer to have them fixed. To avoid diff noise, please don't remove the spaces in the pgsql and mysql $replace ;-)
- when creating the tables, I'd avoid the 0000-00-00 default timestamps. At least for MySQL, they caused some problems in strict mode, therefore we'll use 2000-01-01 in the future. It's probably a good idea to also do this in sqlite. (Hint: just use `{DATE}` or `{DATECURRENT}` which will use the value defined in $replace.)
- as a general note, use the `{WHATEVER}` defined in $replace whenever possible - that makes the queries more readable and is also a good typo prevention. As a side effect, it also ensures you filled $replace correctly ;-)
- in the fetchmail table,
    - `id` should be auto_increment
    - `src_auth` should ideally be enum (is that available in sqlite?), otherwise varchar or text are also ok
- in the quota2 table, `username` should be varchar(255). Actually you uncovered a bug, it's also wrong in mysql and pgsql ;-)
- your CREATE TABLE statements don't specify the charset. Is this an oversight or doesn't sqlite need a charset specified? If so, what will it use?

I have to admit I never used sqlite, but you probably already guessed that from some of the questions above ;-)

b) some nitpicking about the functions.inc.php changes:
- introducing the db_sqlite() function is a good idea, but you should use it everywhere ;-) (yes, I know we have lots of $CONF['database_type'] and Config::Read('database_type') checks, but I want to get rid of them instead of adding new ones ;-)  (please do that only for the sqlite parts, the cleanup for mysql and postgresql should be a separate change)
- a comment about the differences in db_quota_text() ("concat" vs. "||") would be helpful
- db_insert() and db_update(): I know `... = (db_sqlite() ? ... : ...)` is shorter, but I'd prefer a more readable if / else

c) a funny note about the *.lang changes:
- the comment about dateformat_sqlite talks about MySQL ;-) (BTW: if sqlite uses the same format string syntax as MySQL, you can just use the dataformat_mysql instead of adding dateformat_sqlite)

I hope this review doesn't shock you too much ;-)

Like David, I only read the patch, but didn't test it.

palepurple:
upgrade.php : "..... traditionally, the function numbers correspond to the SVN revision on sourceforge, but we might drop that tradition with the changed upgrade handling. ...."

Which : 
 * doesn't work for git when we join the rest of the world and leave sf.net
 * was a pain to keep in sync with the svn revision anyway
 * and was removed a year ago (svn r1740 on my uptime anniversary, 2015-02-17)

" stop relying on subversion keyword for database upgrades; just iterate over all user defined functions and find the largest numeric one"

So I'm happy with any incremental numbers / function names.

phyrog:
Thanks for the input. I'll fix the points you mentioned.

I named the upgrade function with the same revision number, because the database schema has not changed since 1795 (more like "this revision introduced these changes to the schema").

phyrog:
I pushed a newer version with two TODOs (see last comment). I think I fixed everything else you mentioned.

SQLite uses a database wide encoding (UTF8, it's now explicit).

cboltz:
Thanks for updating the patch!

Remaining questions and notes:
- for create_page_browser(), a quick workaround could be to display page numbers (1, 2, 3, ...) instead of the range (a-c, d-f, ...). Or just loop over all lines (which you need for counting them anyway IIRC) and pick the ones you need.
- when/how often is `PRAGMA encoding = "UTF-8"` needed? (Depending on the answer, it would be good to move it to db_query_parsed() in setup.php or even to db_connect() in functions.inc.php.)
    - once when creating the tables
    - every time when modifying the table layout
    - every time when reading or writing data from/to the database?
- setup.php: 
    - {FULLTEXT} for the fetchmail.src_auth looks too big. It will only contain one keyword, so a VARCHAR should be more than enough
    - the "fast forward" is still inside the loop - is that intentional?
    - does sqlite support "named" keys like "PRIMARY KEY on_vacation (\`on_vacation\`, \`notified\`)"? If yes, it's a good idea to use them because that makes it easier if we ever need to change or delete a key.

phyrog:
I'm trying to use common table expressions for create_page_browser. It looks like it could be working, but I it needs a little more tweaking. The PRAGMA needs to be run once before any data is entered into the database, so I think running it in the setup would be appropriate.
 - setup.php:
  - will change it, but it doesn't really matter, since sqlite does not differentiate between text and varchar. It actually does not really care about types at all. You could insert a string into an int column.
  - the fast forward inside the loop is intentional, since there may be later upgrade functions that need to be run
  - I don't think it supports this (https://www.sqlite.org/lang_createtable.html#x908)

phyrog:
Alright, `create_page_browser` is working now, but I had to rewrite the calculation of affected/returned rows in `db_query` quite a bit. The query in `create_page_browser` begins with `WITH`, but is a `SELECT` query. I could have extended the regex to also match queries starting with `WITH`, but that feels like a hack and could also cause future errors (most - if not all - other queries may also start with `WITH`).

So now it just gets both the `_num_rows` and `_affected_rows` and returns the bigger one (`_affected_rows` should be 0 on `SELECT`s and `_num_rows` should be 0 on modifying queries, but might throw a warning).
There might be an edge case when you update a table while querying another one, but I don't think that such a query is used in this project.

cboltz:
A quick (and evil ;-)) test shows you'll need to improve the error handling ;-)

I set $CONF['database_type'] = 'sqlite', but did NOT set $CONF['database_path']. Result:

* PHP Notice:  Undefined index: database_path in .../functions.inc.php on line 1268
* PHP Fatal error:  Uncaught exception 'Exception' with message 'Unable to expand filepath' in .../functions.inc.php:1268\nStack trace:\n#0 .../functions.inc.php(1268): SQLite3->__construct('')\n#1 .../setup.php(237): db_connect(true)\n#2 {main}\n  thrown in .../functions.inc.php on line 1268

I'd recommend to at least test if $CONF[database_path] is non-empty. Ideally you should check if dirname($CONF[database_path]) exists, is a directory and is writeable.

Also, I'd recommend to use a separate directory for the database (for example "/etc/postfix/sqlite/postfixadmin.db") for better security - the webserver will need write access to the database and also temporarily creates "/etc/postfix/sqlite/postfixadmin.db-journal". That means it needs write permissions for the directory, and I do NOT want write permissions for apache in /etc/postfix ;-)  I know this is user-configurable, but the config example should include a separate directory.

BTW - is $CONF[database_name] used for sqlite at all? Maybe you should use that to avoid yet another config variable ;-)

.

Another bug I just noticed: in the domain list, I see "4 / aliases" and "1 / mailboxes" instead of "4 / 100" and "1 / 50". It seems your query in db_quota_text includes the field name instead of its value. You can fix this by changing the line to:

    ELSE (coalesce($count,0) || ' / ' || $quota)
                                    ^^^^^^^^^^^

.

For db_query, I don't really like the idea to do "superfluous" work for mysql and postgresql. Instead, what about this?

     if ($error_text == "") {
        if (db_sqlite()) {
            # do whatever you need
        } elif (preg_match("/^SELECT/i", trim($query))) {

.

And a note about fast forward in setup.php again ;-)

Basically you do:

    $current_version = 0;   # that's what you get if the database doesn't exist yet
    
    for ($i = $current_version +1; $i <= $target_version; $i++) {
        # ...
        } elseif(db_sqlite()) {
            # ...
            if ($i < 1824) {
              // Fast forward to the first revision supporting SQLite
              $i = 1823;
              continue;
            }
    }

So the `if ($i < 1824)` check gets done in every round of the loop, but for obvious reasons it only changes $i once - in the first run of the loop.

You'll get the same behaviour with

    if (db_sqlite() && $current_version < 1824) {
        # Fast forward to the first revision supporting SQLite
        $current_version = 1823;
    }
    for ($i = $current_version +1; $i <= $target_version; $i++) {

.

Trying to create a fetchmail job, I get

* Invalid query: datatype mismatch, referer: http://postfixadmin.tux/HEAD-clean/edit.php?table=fetchmail
* caused by query: INSERT INTO prefix_fetchmail (id,domain,mailbox,src_server,src_auth,src_user,src_password,src_folder,poll_time,fetchall,keep,protocol,usessl,sslcertck,sslcertpath,sslfingerprint,extra_options,mda,date,returned_text,active,created,modified) VALUES ('','cboltz.de','info@cboltz.de','cboltz.de','password','info','dG9wc2VjcmV0','','10','0','0','POP3','0','0','','','','','1','','1',datetime('now'),datetime('now')), referer: http://postfixadmin.tux/HEAD-clean/edit.php?table=fetchmail

Can you please check what's wrong here?
(dear spammers - the query contains example data, nothing that actually works ;-))

.

Finally some good news - everything else I tested works :-)

phyrog:
Alright, I believe I fixed everything you mentioned now. The SQL problem was caused by inserting a row with `id` = `''`. These are now removed before inserting.

> Finally some good news - everything else I tested works :-)

Sounds good :)

cboltz:
The `id` field needs to be fixed/removed from the query elsewhere ;-) (db_insert() is quite low-level and shouldn't mess with the query)

Please drop the change in functions.inc.php db_insert() - the fix has to happen in PFAHandler set():

```
    public function set($values) {
       [...]

-       if ($this->new == 1) {
+       if ($this->new == 1 && $this->struct[$this->id_field]['editable']) {
            $values[$this->id_field] = $this->id;
        }
```

I wrote this code before I wrote FetchmailHandler, which might explain why I expected the id_field to be always editable when creating a new item. Interestingly, mysql doesn't complain about including the `id` field in the query, and for postgresql at least I didn't receive a bugreport ;-)

I didn't test this change, but it should work ;-) and also shouldn't break other *Handler classes because they set the field editable if $this->new. Please give it a quick test and update the pull request.

The other changes look good, so this is hopefully the last thing to fix :-)

phyrog:
Unfortunately that didn't fix the error, but setting `dont_write_to_db` for `id` to `1` did. It also uncovered a bug in the row-counting part of `db_query`, which [ran the query twice](https://bugs.php.net/bug.php?id=64531) when using `fetchArray` after inserting, updating or deleting something. Fixed this now too.

cboltz:
I'm surprised that the change in PFAHandler set() didn't work - I'll have to test this myself one day ;-) but the solution you used will fix it for now.

Also, thanks for solving the duplicate execution - it's always nice[tm] if you find bugs in PHP itsself ;-)

I just tested with your latest patch, and everything looks good.

David, please merge this pull request and mirror it back to SVN ;-)

PR #11, opened 2016-03-08
Adding an option to configure the hostname in HELO when sending email from Postfix Admin

captainark:
Hi,

I was having problems when sending mail from Postfix Admin because the php function php_uname("n") was returning my server's shortname instead of it's FQDN.

This might not be the best way to go about solving this, but since it works for me, I figured I'd do a pull request.


palepurple:
Hi - Thanks for the patch.

I think I'd rather something along the lines of leaving the existing php_uname('n') behaviour there, but if the config setting is present and non-empty, then it's used instead.

e.g.
```
$smtp_client = php_uname('n');
if(!empty($config['smtp_client'])) {
    $smtp_client = $config['smtp_client'];
}
```

captainark:
Thank you for your feedback.

I just pushed a new version ; let me know what you think !

palepurple:
Thanks!

PR #12, opened 2016-03-14
Add support for multiple servers

Janfred:
Add support for multiple servers.

Useful for hot standby or backup systems with replicated databases.

The main problem is that, if you want to have a backup mx, you need to copy all mailbox setting changes you make on the first system to the second one.

With this new feature you can just replicate your database and use it for both, the primary and the secondary mailserver.

You can also do load balancing by sending one domain to the first and an other domain to the second mailserver as primary.


palepurple:
Thank you for taking the time to submit the pull request. I've had a quick eye-ball of it, and nothing immediate is jumping out.
(The relay_recipient_maps is a useful addition too - as I've found myself needing this too)

Janfred:
There is one small todo, i don't have any idea how to solve it:
At the server list is a select field for the admin. I could not remove that like in the admin list.
Maybe you can give me a hint how to remove that.

DOCUMENTS/POSTFIX_CONF.txt
```diff
@@ -13,12 +13,19 @@ postconf -m)  Three main.cf variables are involved:
 
 virtual_mailbox_domains = proxy:mysql:/etc/postfix/sql/mysql_virtual_domains_maps.cf
 virtual_alias_maps =
-   proxy:mysql:/etc/postfix/sql/mysql_virtual_alias_maps.cf,
```
cboltz: There is a reason why mysql_virtual_alias_maps.cf is listed first - an alias domain can "override" some aliases. To keep this working, mysql_virtual_alias_maps.cf needs to be the first in the list.

Why did you move it around?

DOCUMENTS/POSTFIX_CONF.txt
```diff
@@ -13,12 +13,19 @@ postconf -m)  Three main.cf variables are involved:
...
 virtual_mailbox_maps =
    proxy:mysql:/etc/postfix/sql/mysql_virtual_mailbox_maps.cf,
    proxy:mysql:/etc/postfix/sql/mysql_virtual_alias_domain_mailbox_maps.cf
+transport_maps = proxy:mysql:/etc/postfix/sql/mysql_transport_maps.cf
+relay_domains = proxy:mysql:/etc/postfix/sql/mysql_relay_domains_maps.cf
+relay_recipient_maps =
+   proxy:mysql:/etc/postfix/sql/mysql_virtual_alias_domain_maps.cf,
+   proxy:mysql:/etc/postfix/sql/mysql_virtual_alias_domain_catchall_maps.cf,
+   proxy:mysql:/etc/postfix/sql/mysql_virtual_alias_maps.cf,
+   proxy:mysql:/etc/postfix/sql/mysql_backup_mx_recipients.cf
```
cboltz: Same question/note here - IMHO mysql_virtual_alias_maps.cf needs to be first in the list.

DOCUMENTS/POSTFIX_CONF.txt
```diff
@@ -90,12 +97,33 @@ user = postfix
...
 #expansion_limit = 100
 
+mysql_relay_domains_maps.cf:
+user = postfix
+password = password
+hosts = localhost
+dbname = postfix
+query = SELECT domain FROM domain WHERE domain='%s' AND active = '1' AND primarymx !='$myhostname' AND primarymx!=''
+
+mysql_backup_mx_recipients.cf
```
cboltz: minor nitpick: the ':' at the end of the line is missing (which will break the scripting part of POSTFIX_CONF.TXT)

DOCUMENTS/POSTFIX_CONF.txt
```diff
@@ -167,6 +199,7 @@ for file in $map_files ; do
 			s/^password =.*/password = $password/ ;
 			s/^hosts =.*/hosts = $hosts/ ;
 			s/^dbname =.*/dbname = $dbname/ ;
+            s/\$myhostname/$myhostname/ ;
```
cboltz: whitespace looks wrong - tab vs. spaces?

functions.inc.php
```diff
@@ -654,7 +654,26 @@ function list_domains () {
...
+// list_servers
+// Action: List all available servers.
+// Call: list_servers ()
+//
+function list_servers () {
+    $list = array();
 
+    $table_server = table_by_key('server');
+    $list[null] = "none";
+    $result = db_query ("SELECT server FROM $table_server ORDER BY server");
```
cboltz: Please use ServerHandler->getList() instead of "raw" SQL queries - see list_admins() for an example.

Since you only call this function in PFAHandler, you might as well embed the 3 lines of code there - functions.inc.php is already big enough ;-)

model/PFAHandler.php
```diff
@@ -59,6 +59,10 @@
     # set in __construct()
     protected $allowed_domains = false;
 
+    # if $server_field is set, this is an array with the server list
```
cboltz: You mention $server_field, but it isn't set anywhere...

model/ServerHandler.php
```diff
@@ -0,0 +1,167 @@
...
+/**
+ *
+ */
+class ServerHandler extends PFAHandler{
+
+    protected $db_table = 'server';
+    protected $id_field = 'server';
+
+    protected function validate_new_id() {
+        return true;
```
cboltz: I'm not sure if we want or need to enforce a full host name, but I'd recommend to at least check for "not empty" and "not only whitespace" ;-)

model/ServerHandler.php
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    protected function no_domain_field() {
+        # PFAHandler die()s if domain field is not set. Disable this behaviour for AdminHandler.
```
cboltz: "... for AdminHandler" - wrong comment for this file ;-)

model/ServerHandler.php
```diff
@@ -0,0 +1,167 @@
...
+            '_can_delete' => pacol( 0,          0,      1,      'int',  '',              '', 0,
+              /*options*/ '',
+              /*not_in_db*/ 0,
+              /*dont_write_to_db*/ 1,
+              /*select*/ $this->is_superadmin . ' as _can_delete' ),
+        );
+    }
+
+    protected function initMsg() {
+        $this->msg['error_already_exists'] = 'pAdminCreate_server_server_text_error';
```
cboltz: Feel free to use something like 'server_already_exists' as message name - the pAdmin* message names are only there for historical reasons[tm]. I prefer names that relate more to the message content, and the only reason why the pAdmin* messages still have this name is that nobody renamed them.

Anyway, there's no reason to use hard to understand pAdmin* message names for new messages.

This comment also applies to the other pAdmin* messages you added in this function and in webformConfig().

model/ServerHandler.php
```diff
@@ -0,0 +1,167 @@
...
+
+        $cmdarg1=escapeshellarg($this->id);
+        $command= "$script $cmdarg1";
+        $retval=0;
+        $output=array();
+        $firstline='';
+        $firstline=exec($command,$output,$retval);
+        if (0!=$retval) {
+            error_log("Running $command yielded return value=$retval, first line of output=$firstline");
+            $this->errormsg[] = 'Problems running server postcreation script!';
```
cboltz: hardcoded/non-translatable error message

Also, you set a (translatable) error message in storemore(), which will probably result in two error messages being displayed.

model/ServerHandler.php
```diff
@@ -0,0 +1,167 @@
...
+
+    protected function server_postdeletion() {
+        $script=Config::read('server_postdeletion_script');
+
+        if(empty($script)) {
+            return true;
+        }
+
+        if (empty($this->id)) {
+        }
```
cboltz: no reaction if $this->id is empty?

I'd tend to error out.

BTW: Just curious - do you have a real usecase for server postcreate/postdelete scripts, or did you implement the hooks just in case someone needs them?

model/ServerHandler.php
```diff
@@ -0,0 +1,167 @@
...
+
+        $cmdarg1=escapeshellarg($this->id);
+        $command= "$script $cmdarg1";
+        $retval=0;
+        $output=array();
+        $firstline='';
+        $firstline=exec($command,$output,$retval);
+        if (0!=$retval) {
+            error_log("Running $command yielded return value=$retval, first line of output=$firstline");
+            $this->errormsg[] = 'Problems running server postdeletion scripts!';
```
cboltz: another hardcoded/non-translatable error message

model/ServerHandler.php
```diff
@@ -0,0 +1,167 @@
...
+            return false;
+        }
+        
+        if ( ! $this->view() ) {
+            $this->errormsg[] = Config::Lang('server_does_not_exist');
+            return false;
+        }
+
+        $handler = new DomainHandler(0, $this->admin_username);
+        $handler->getList("primarymx = '" . escape_string($this->id) . "'");
```
cboltz: You can use an array as getList parameter:

    $handler->getList( array("primarymx" => $this->id) );

Note that this will enable some additional checks in getList, but that (hopefully) shouln't hurt.

upgrade.php
```diff
@@ -1623,7 +1622,53 @@ function upgrade_1824_sqlite() {
...
+function upgrade_1825_sqlite() {
+    $server = table_by_key('server');
+    db_query_parsed("
+        CREATE TABLE $server (
+         `server` varchar(100) NOT NULL,
+         `description` varchar(255) NOT NULL,
+         `address` varchar(100) NOT NULL,
+         `created` {DATE},
+         `modified` {DATECURRENT},
+         {PRIMARY} (`email`));
```
cboltz: I'd prefer if you can create one function for all database types. (Add additional {...} markers if needed.)

This will also avoid problems like the {PRIMARY} (`email`) which refers to a non-existing field ;-)

Also, description should probably be {UTF8}.

upgrade.php
```diff
@@ -1623,7 +1622,53 @@ function upgrade_1824_sqlite() {
...
+    ");
+}
+
+function upgrade_1826_mysql() {
```
cboltz: Please merge these 3 functions into one that calls _db_add_field() ;-)

DOCUMENTS/POSTFIX_CONF.txt
Janfred: The reason i remember was that recursive lookups failed.
e.g. some_mail@host aliases to webmaster@host
webmaster@host aliases to my_name@host

In the default settings this lookup failed.
Could you name a setting where this order of handling is needed?

Janfred:
Ok, i've done several changes:
* The language keys are changed.
* The list_servers method 'moved' to the DomainHandler. I don't know if this is the best coding style, but it is better placed then in the functions.inc.php
* The 2nd Migration is merged into one. ( Just one small question: does this already has effect with the sqlite-Implementation from phyrog? )
* The primary key is set to the correct one (copy-paste-error)
* The primary key is now checked that it is not empty on creation

The postcreation and postdeletion hook is only there because i followed the implementation of domains. I don't know if there is a need for server-post(creation|deletion)-scripts.

I still dont get what problem the changed order in the virtual_alias_maps would cause.
If you could tell me some usecases, I could check. if there is a solution to allow recursive lookups as well.

DOCUMENTS/POSTFIX_CONF.txt
cboltz: Hmm, I'm surprised that this fails because IIRC I have more than one "chained" alias ;-) - and I'm even more surprised that changing the order fixed it for you.

The setting that will probably break is:
* alias a@host -> foo@example.com
* alias a@otherhost -> bar@example.com
* alias b@otherhost -> b@example.com (and some other aliases that don't exist @host)
* alias domain host -> otherhost

The goal is: all mails that are sent to @host should be forwarded to the same address @otherhost (if this address exists) - except a@ because the a@host alias overrides this.

Janfred:
For the postfix-issue: I checked it with my local setup. You were right. I don't know what went wrong in my normal setup...
I changed this in the `POSTFIX_CONF.txt`

Janfred:
I did one final change: The server tab is now only visible for super-admins.

I thought of making only servers visible to non-global admins which are primary servers for their domain(s).
But I don't think this is very important to look at for non-global admins.

I left the primary_mx column in the domain overview, maybe this is useful. If it is not, i'll remove it too.

model/PFAHandler.php
```diff
@@ -59,6 +59,9 @@
     # set in __construct()
     protected $allowed_domains = false;
 
+    # array with the server list set in __construct()
+    protected $allowed_servers = false;
+
```
cboltz: (sorry for the long delay!)

This is only used in DomainHandler, so it should be defined there. (If it's used by more *Handler classes in the future, PFAHandler would be ok.)

upgrade.php
```diff
@@ -1623,7 +1622,39 @@ function upgrade_1824_sqlite() {
...
+         `address` varchar(100) NOT NULL,
+         `created` {DATE},
+         `modified` {DATECURRENT},
+         {PRIMARY} (`server`));
+    ");
+}
 
+function upgrade_1826() {
+    $domain = table_by_key('domain');
+    _db_add_field($domain, 'primarymx','VARCHAR(100) {LATIN1} NOT NULL','transport');
```
cboltz: _db_add_field calls table_by_key() itsself, which means we run into double-mapping of the name. 
Just call _db_add_field('domain', ...).

templates/menu.tpl
```diff
@@ -21,6 +21,17 @@
 {else}
 	<li><a target="_top" href="{#url_main#}">{$PALANG.pMenu_main}</a></li>
 {/if}
+{* list-server *}
+{if $authentication_has_role.global_admin}
+{strip}
+    <li><a target="_top" href="{#url_list_server#}">{$PALANG.pAdminMenu_list_server}</a>
```
cboltz: Not everybody will need support for multiple servers, so it would be nice to have a config option (maybe $CONF['multiple_servers']) that can disable this feature.

If disabled,
- the menu item should be hidden
- ServerHandler should refuse any edits, which means:
  - set _can_edit and _can_delete to Config::bool('multiple_servers') && $this->is_superadmin) in initStruct()
  - also add similar checks for can_create (in initMsg()) and delete()
- DomainHandler should not display the server field if multiple server handling is disabled

DOCUMENTS/POSTFIX_CONF.txt
```diff
@@ -90,12 +98,41 @@ user = postfix
...
 #optional query to use when relaying for backup MX
 #query           = SELECT domain FROM domain WHERE domain='%s' AND backupmx = '0' AND active = '1'
 #expansion_limit = 100
 
+mysql_relay_domains_maps.cf:
+user = postfix
+password = password
+hosts = localhost
+dbname = postfix
+query = SELECT domain FROM domain WHERE domain='%s' AND active = '1' AND primarymx !='$myhostname' AND primarymx!=''
```
cboltz: We also have the "backup MX" field (which isn't covered by the existing POSTFIX_CONF.TXT).
Would it be possible to include it in your query?

DOCUMENTS/POSTFIX_CONF.txt
```diff
@@ -90,12 +98,41 @@ user = postfix
...
+hosts = localhost
+dbname = postfix
+query = SELECT CONCAT('%u', '@', '%d') AS recipient FROM domain WHERE domain.domain = '%d' AND domain.backupmx = 1
+
+mysql_transport_maps.cf:
+user = postfix
+password = password
+hosts = localhost
+dbname = postfix
+query = SELECT address FROM server JOIN domain ON domain.primarymx = server.server WHERE domain.domain = '%s' OR CONCAT('.', domain.domain) = '%s'
```
cboltz: We also have the "transport" field, which would be a natural fit for transport_maps. POSTFIX_CONF.TXT doesn't have example config for it yet, but with a mysql_transport_maps.cf added, it should probably be handled there ;-)

DOCUMENTS/POSTFIX_CONF.txt
```diff
@@ -167,6 +208,7 @@ for file in $map_files ; do
 			s/^password =.*/password = $password/ ;
 			s/^hosts =.*/hosts = $hosts/ ;
 			s/^dbname =.*/dbname = $dbname/ ;
+			s/\$myhostname/$myhostname/ ;
```
cboltz: If you use @@myhostname@@ in the config file template, you make your life easier (no need for escaping the $ sign) and also make it clear for people who read POSTFIX_CONF.TXT that it's only a placeholder.

BTW: A note about being @@myhostname@@ (or $myhostname if you insist on the name) being a placeholder would be nice ;-)

model/ServerHandler.php
```diff
@@ -0,0 +1,173 @@
...
+    protected $db_table = 'server';
+    protected $id_field = 'server';
+
+    protected function validate_new_id() {
+        if ($this->id == null || $this->id == "") {
+            $this->errormsg[$this->id_field] = Config::lang('server_must_not_be_empty');
+            return false;
+        }
+        else
+        {
```
cboltz: nitpicking: please merge `} else {` into one line

DOCUMENTS/POSTFIX_CONF.txt
```diff
@@ -90,12 +98,41 @@ user = postfix
...
+hosts = localhost
+dbname = postfix
+query = SELECT CONCAT('%u', '@', '%d') AS recipient FROM domain WHERE (domain.domain = '%d' OR '%d' LIKE CONCAT('%%.', domain.domain)) AND domain.backupmx = 1
+
+mysql_transport_maps.cf:
+user = postfix
+password = password
+hosts = localhost
+dbname = postfix
+query = SELECT address FROM server JOIN domain ON domain.primarymx = server.server WHERE (domain.domain = '%s' OR CONCAT('.', domain.domain) = '%s') AND domain.primarymx!='@@myhostname@@'
```
tumd: Shouldn't the 'transport_maps'-query also include ` AND active = '1'`?

FWIW, I love this pull request, and are working on implementing it in my own mail-server setup.

Janfred:
I've rebased all changes to comply with the new master status.

You might want to recheck my implementations if they comply with your code style.

I haven't done the adjustment @cboltz suggested to use the already available 'transport' field.  
The transport is not the same for the primary and secondary server. If the secondary server receives a mail for the primary it will be a relay transport. For the primary it will be (depending on the overall setup) a virtual or local transport.

config.inc.php
```diff
@@ -523,6 +528,22 @@ function x_struct_admin_modify($struct) {
 // $CONF['domain_postdeletion_script']='sudo -u courier /usr/local/bin/postfixadmin-domain-postdeletion.sh';
 $CONF['domain_postdeletion_script'] = '';
 
+// Optional
+// Script to run after creation of servers.
+// Not that this may fail if PHP is nun in "safe mode", or if
```
jdsn: `Not` or `Note` ?

config.inc.php
```diff
@@ -523,6 +528,22 @@ function x_struct_admin_modify($struct) {
...
+// operating system features (such es SELinux) or limitations
+// prevent the web-server from executing external scripts.
+// Parameters: (1) server
+$CONF['server_postcreation_script'] = '';
+
+// Optional
+// Script to run after deletion of servers.
+// Not that this may fail if PHP is nun in "safe mode", or if
```
jdsn: `Not` or `Note` ?

languages/bg.lang
```diff
@@ -379,6 +383,26 @@ $PALANG['pFetchmail_desc_date']             = 'Date of last polling/configuratio
...
+$PALANG['pAdminMenu_create_server'] = 'New Server'; # XXX
+$PALANG['server'] = 'Server'; # XXX
+$PALANG['smtpaddress'] = 'SMTP-Address'; # XXX
+$PALANG['server_address_description'] = 'smtp address of the remote server for relaying (e.g. smtp:[1.2.3.4]:25 )'; # XXX
+$PALANG['server_creation'] = 'Add a server'; # XXX
+$PALANG['server_edit'] = 'Edit a server'; # XXX
+$PALANG['server_button'] = 'Add server'; # XXX
+$PALANG['server_create_error'] = 'Creating server %s failed'; # XXX
+$PALANG['server_edit_error'] = 'Modifying server %s failed'; # XXX
+$PALANG['server_store_success'] = 'The Server %s has successfully been added.'; # XXX
```
jdsn: just nitpicking: why is `Server` uppercase here?

languages/bg.lang
```diff
@@ -379,6 +383,26 @@ $PALANG['pFetchmail_desc_date']             = 'Date of last polling/configuratio
...
+$PALANG['server_already_exists'] = 'The server already exists'; # XXX
+$PALANG['server_does_not_exist'] = 'The server %s does not exist.'; # XXX
+$PALANG['confirm_delete_server'] = 'Do you really want to delete the server %s?'; # XXX
+$PALANG['delete_server_domain_target'] = 'There are still domains, where this server is the primary one. Change them first'; # XXX
+$PALANG['server_updated'] = 'The Server %s was updated'; # XXX
```
jdsn: just nitpicking again :)

Janfred:
I have updated this pull request to Postfixadmin v3.
There is still work to do, but I'd like to hear if this goes in the right direction @cboltz 

I haven't updated the language files yet, nor the postfix configuration script.

model/ServerHandler.php
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        $cmdarg1=escapeshellarg($this->id);
+        $command= "$script $cmdarg1";
+        $retval=0;
+        $output=array();
+        $firstline='';
+        $firstline=exec($command,$output,$retval);
+        if (0!=$retval) {
+            error_log("Running $command yielded return value=$retval, first line of output=$firstline");
```
DavidGoodwin, line 114: Why not (for example) just json_encode($output) at the end of this message, rather than restricting it to the first line ?

DavidGoodwin:
....
> 
> ....nor the postfix configuration script.


By that - do you mean whatever config is needed for Postfix to use this functionality ?

model/ServerHandler.php
Janfred: To be honest, I've just copied this piece of code from [model/DomainHandler.php#L292](https://github.com/postfixadmin/postfixadmin/blob/e8f27969a3a1be7461efc3ece169605fec4c5f43/model/DomainHandler.php#L292)

I could change that in my pull request, but I guess it should then be changed in all model handlers.

Janfred:
> By that - do you mean whatever config is needed for Postfix to use this functionality ?

In the previous version of my pull request I changed the POSTFIX_CONF.txt to support the multiple server feature. But I think my changes were not compatible with the standard-setup.
I'd be happy about any idea how to address this issue.

DavidGoodwin:
Any chance you could rebase your branch onto master and add some basic documentation - e..g what changes are needed to the database queries postfix uses for this to be in a usable state to merge in?

Neustradamus:
@Janfred: It is possible to rebase your PR?

Janfred:
I'll add it to my todo-List for when I have a bit time to spare.

ETE-Design:
@Janfred Did you manage to rebase your PR? Would be a really nice feature to have :-)

PR #18, opened 2016-10-19
#75 Enable users to reset their passwords

stissot:
Enables users and admins to reset their forgotten password, by receiving an OTP by email to an alternative address and/or by SMS to their mobile phone.

See also original request: https://sourceforge.net/p/postfixadmin/feature-requests/75/

**Implementation**
Below the login form for users and/or admins, a 'reset password' link is displayed. Once clicked, an informative page explains the password reset process to the user.
A confirmation message is sent to the user's email address and/or mobile phone on file. This message contain an URL with a reset code (computed as a hash of request date + user name + secret salt). When clicking the URL, the user is forwarded to form where, if the reset code matches, the user can then set a new password by himself. If the unique ID doesn't match, an error page is displayed. This feature can be turned on/off separately for admins and users.
The [Clickatell API](https://www.clickatell.com/developers/api-docs/http-using-the-http-api/) is used to send SMS, you need to open a Clickatell account, purchase some credits, create a HTTP API and fill the API credentials in Postfixadmin's config file. 2 new fields "phone" and "email_other" are added by the upgrade script to the mailbox and admin tables.

**Future improvement**
Limit the number of password reset requests that can be done by user/IP address.

postfixadmin_password_reset

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -119,6 +119,14 @@ $CONF['database_tables'] = array (
 // Leave blank to send email from the logged-in Admin's Email address.
 $CONF['admin_email'] = '';
 
+// Site admin name
+// This will be used as signature in notification messages
+$CONF['admin_name'] = 'Postmaster';
+
+// Site admin phone number
+// This will be used if a user cannot access his/her email and needs support
+$CONF['admin_phone'] = '';
+
 // Mail Server
 // Hostname (FQDN) of your mail server.
 // This is used to send email to Postfix in order to create mailboxes.
@@ -572,6 +580,18 @@ $CONF['new_quota_table'] = 'YES';
 // http://www.php.net/manual/en/function.imap-open.php
 $CONF['create_mailbox_subdirs_hostoptions'] = array();
 
+// Optional:
+// Allows a user to reset his forgotten password with a code sent by email/SMS
+$CONF['forgotten_user_password_reset'] = true;
+// Allows an admin to reset his forgotten password with a code sent by email/SMS
+$CONF['forgotten_admin_password_reset'] = true;
+
+// Clickatell gateway to send SMS code for password reset
+// API type: HTTP
+$CONF['clickatell_api_id'] = '';
+$CONF['clickatell_user'] = '';
+$CONF['clickatell_password'] = '';
+$CONF['clickatell_sender'] = '';
 
 // Theme Config
 // Specify your own logo and CSS file
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -1939,4 +1939,21 @@ function getRemoteAddr() {
     return $REMOTE_ADDR;
 }
 
+
+/**
+ * Returns a hash for a username valid for one day
+ *
+ * @param String $username user name
+ * @return String password recovery code
+ */
+function getPasswordRecoveryCode($username)
+{
+    $username = trim(strtolower($username));
+    $date = date('Y-m-d');
+
+    $code = substr(strtoupper(md5('SECRET SALTING PHRASE' . $username . $date)), 0, 6);
+
+    return $code;
+}
+
 /* vim: set expandtab softtabstop=4 tabstop=4 shiftwidth=4: */
--- a/languages/en.lang
+++ b/languages/en.lang
@@ -146,6 +146,10 @@ $PALANG['pCreate_mailbox_username_text_error1'] = 'The EMAIL is not valid!';
 $PALANG['pCreate_mailbox_username_text_error3'] = 'You have reached your limit to create mailboxes!';
 $PALANG['pCreate_mailbox_password_text'] = 'Password for POP3/IMAP';
 $PALANG['pCreate_mailbox_name_text'] = 'Full name';
+$PALANG['pCreate_mailbox_phone'] = 'Mobile phone';
+$PALANG['pCreate_mailbox_phone_desc'] = "Used to send a SMS if the password is forgotten";
+$PALANG['pCreate_mailbox_email'] = 'Other e-mail';
+$PALANG['pCreate_mailbox_email_desc'] = "Used if the password is forgotten";
 $PALANG['pCreate_mailbox_mail'] = 'Send Welcome mail';
 $PALANG['pCreate_mailbox_result_error'] = 'Creating the mailbox %s failed!';
 $PALANG['pCreate_mailbox_result_success'] = 'The mailbox %s has been added to the mailbox table.';
@@ -173,6 +177,16 @@ $PALANG['pPassword_password_text_error'] = 'The passwords that you supplied don\
 $PALANG['change_password'] = 'Change Password';
 $PALANG['pPassword_result_error'] = 'Changing the password for %s failed!';
 $PALANG['pPassword_result_success'] = 'The password for %s has been changed.';
+
+$PALANG['pPassword_recovery_title'] = 'Follow the instructions to reset your password.';
+$PALANG['pPassword_recovery_button'] = 'Send me the code';
+$PALANG['pPassword_recovery_email_body'] = "Hello,\n\nUse the following link to change your email password:\n%s\n\nRegards,\n\n" . $CONF['admin_name'];
+$PALANG['pPassword_recovery_email_sent'] = 'An email was sent to:';
+$PALANG['pPassword_recovery_sms_body'] = "Hello,\nThe code to change your password is: %s\n" . $CONF['admin_name'];
+$PALANG['pPassword_recovery_sms_sent'] = 'An SMS was sent to:';
+$PALANG['pPassword_recovery_no_alternative'] = 'No alternative contact info were found. Please contact the support at ' . $CONF['admin_email'] . 'or by phone to ' . $CONF['admin_phone'];
+$PALANG['pPassword_password_code'] = 'Code sent by email/SMS';
+$PALANG['pPassword_code_text_error'] = 'Invalid code';
 
 $PALANG['pEdit_vacation_set'] = 'Change / Set away message';
 $PALANG['pEdit_vacation_remove'] = 'Remove away message';
@@ -298,6 +312,7 @@ $PALANG['pAdminEdit_admin_result_success'] = 'The admin %s has been modified.';
 $PALANG['pUsersLogin_welcome'] = 'Mailbox users login to change your password and aliases.';
 $PALANG['pUsersLogin_username_incorrect'] = 'Your login is not correct. Make sure that you login with your email address!';
 $PALANG['pUsersLogin_password_incorrect'] = 'Your password is not correct!';
+$PALANG['pUsersLogin_password_recover'] = 'I forgot my password';
 
 $PALANG['pUsersMenu_vacation'] = 'Auto Response';
 $PALANG['pUsersMenu_edit_alias'] = 'Change your forward';
--- a/languages/fr.lang
+++ b/languages/fr.lang
@@ -145,6 +145,10 @@ $PALANG['pCreate_mailbox_username_text_error1'] = 'L\'adresse courriel est inval
 $PALANG['pCreate_mailbox_username_text_error3'] = 'Vous avez atteint le nombre maximum de comptes courriel !';
 $PALANG['pCreate_mailbox_password_text'] = 'Mot de passe pour le compte POP3/IMAP';
 $PALANG['pCreate_mailbox_name_text'] = 'Nom complet';
+$PALANG['pCreate_mailbox_phone'] = 'Téléphone mobile';
+$PALANG['pCreate_mailbox_phone_desc'] = "Utilisé pour l'envoi de SMS en cas d'oubli du mot de passe";
+$PALANG['pCreate_mailbox_email'] = 'E-mail secondaire';
+$PALANG['pCreate_mailbox_email_desc'] = "Utilisé en cas d'oubli du mot de passe";
 $PALANG['pCreate_mailbox_mail'] = 'Envoyer le message de bienvenue';
 $PALANG['pCreate_mailbox_result_error'] = 'Échec de la création du compte courriel %s !';
 $PALANG['pCreate_mailbox_result_success'] = 'Le compte courriel %s a été ajouté à la table des comptes !';
@@ -172,6 +176,17 @@ $PALANG['pPassword_password_text_error'] = 'Le mot de passe fourni ne correspond
 $PALANG['change_password'] = 'Changer le mot de passe';
 $PALANG['pPassword_result_error'] = 'Impossible de changer le mot de passe de %s !';
 $PALANG['pPassword_result_success'] = 'Le mot de passe de %s a été changé !';
+$PALANG['pPassword_recovery_title'] = 'Suivez les instructions pour réinitialiser votre mot de passe.';
+$PALANG['pPassword_recovery_button'] = 'Envoyez-moi le code';
+$PALANG['pPassword_recovery_email_body'] = "Bonjour,\n\nUtilisez le lien suivant pour modifier votre mot de passe :\n%s\n\nSalutations,\n\n" . $CONF['admin_name'];
+$PALANG['pPassword_recovery_email_sent'] = 'Un code a été envoyé à :';
+$PALANG['pPassword_recovery_sms_body'] = "Bonjour,\nLe code pour modifier votre mot de passe: %s\n" . $CONF['admin_name'];
+$PALANG['pPassword_recovery_sms_sent'] = 'Un code a été envoyé par SMS à :';
+$PALANG['pPassword_recovery_no_alternative'] = "Aucun moyen de contact alternatif n'a été trouvé. Contactez le support à " . $CONF['admin_email'] . ' ou par téléphone ' . $CONF['admin_phone'];
+$PALANG['pPassword_password_code'] = 'Code reçu par email/SMS';
+$PALANG['pPassword_code_text_error'] = 'Code invalide';
+
+>>>>>>> Allows a user or admin to reset his/her forgotten password with a code sent by email/SMS #18
 $PALANG['pEdit_vacation_set'] = 'Activer le répondeur';
 $PALANG['pEdit_vacation_remove'] = 'Désactiver le répondeur';
 $PALANG['pVacation_result_error'] = 'Impossible de mettre à jour les réglages du répondeur de %s !';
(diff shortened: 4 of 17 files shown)


languages/fr.lang
```diff
@@ -170,6 +174,17 @@ $PALANG['pPassword_password_text_error'] = 'Le mot de passe fourni ne correspond
 $PALANG['change_password'] = 'Changer le mot de passe';
 $PALANG['pPassword_result_error'] = 'Impossible de changer le mot de passe de %s!';
 $PALANG['pPassword_result_success'] = 'Le mot de passe de %s a été change!';
+
+$PALANG['pPassword_recovery_title'] = 'Suivez les instructions pour réinitialiser votre mot de passe.';
+$PALANG['pPassword_recovery_button'] = 'Envoyer-moi le code';
```
PowerKiKi: Should be `Envoyez-moi`

templates/index.tpl
```diff
@@ -1,7 +1,7 @@
 <!-- {$smarty.template} -->
 {strip}
 {include file="header.tpl"}
-{if $smarty_template != 'login'}
+{if $smarty_template != 'login' && $smarty_template != 'password-recover' && $smarty_template != 'password-change'}
```
PowerKiKi: Should use strict comparaison `!==` when comparing to string

templates/password-change.tpl
```diff
@@ -0,0 +1,28 @@
+<div id="edit_form">
+<form name="mailbox" method="post">
+<table>
+   <tr>
```
PowerKiKi: This file should use tabs instead of spaces for indentations to be consistent with other templates.

users/password-change.php
```diff
@@ -0,0 +1,104 @@
...
+ * Postfix Admin 
+ * 
+ * LICENSE 
+ * This source file is subject to the GPL license that is bundled with  
+ * this package in the file LICENSE.TXT. 
+ * 
+ * Further details on the project are available at : 
+ *     http://www.postfixadmin.com or http://postfixadmin.sf.net 
+ * 
+ * @version $Id: password.php 679 2009-06-30 12:56:30Z GingerDog $
```
PowerKiKi: I think it should read:
```
@version $Id$
```

users/password-change.php
```diff
@@ -0,0 +1,104 @@
...
+ * This source file is subject to the GPL license that is bundled with  
+ * this package in the file LICENSE.TXT. 
+ * 
+ * Further details on the project are available at : 
+ *     http://www.postfixadmin.com or http://postfixadmin.sf.net 
+ * 
+ * @version $Id: password.php 679 2009-06-30 12:56:30Z GingerDog $ 
+ * @license GNU GPL v2 or later. 
+ * 
+ * File: password.php
```
PowerKiKi: copy/paste mistake

users/password-change.php
```diff
@@ -0,0 +1,104 @@
...
+ * Used by users to change their mailbox (and login) password.
+ * Template File: users_password.php
```
PowerKiKi: copy/paste mistake (bis)

stissot:
Thanks @PowerKiKi, this has been fixed

DavidGoodwin:
Thanks for taking the time to submit this. I like the idea of it, but I've not had time to look at your code yet.

functions.inc.php
```diff
@@ -1918,4 +1918,21 @@ function getRemoteAddr() {
...
+ *
+ * @param String $username user name
+ * @return String password recovery code
+ */
+function getPasswordRecoveryCode($username)
+{
+    $username = trim(strtolower($username));
+    $date = date('Y-m-d');
+
+    $code = substr(strtoupper(md5('SECRET SALTING PHRASE' . $username . $date)), 0, 6);
```
cboltz: Using a fixed string + the mail address + the date (even without time) makes the recovery code easily computable and would directly translate into a CVE once we accept this code. (In other words: if you already run this code on a server already, disable it _now_.)

You'll need to add some randomness here, which also means you'll need to store the recovery code and the "valid until" date/time in the database.

Ideally we could use random_bytes(), but it was only added in PHP 7. So you'll need an `if function_exists('random_bytes')' switch with a fallback to openssl_random_pseudo_bytes(). I'd include at least 8 bytes of randomness. If none of these functions exists, throw an error. (Checking for them in setup.php might be a good idea, but it should be a warning only since it only affects an optional feature.)

Also, please include time() in the hash (not really a security improvement, but it can't hurt), and use sha1() instead of md5().

Finally, 6 chars are too short for a password recovery code - with the limited amount of characters used in a md5 or sha1 sum, it's too easy to guess. Please _at least_ make it 10 chars long - longer is better, and if someone forgets his password, he should at least suffer a bit when typing the code ;-)

templates/login.tpl
```diff
@@ -14,7 +14,8 @@
 	</tr>
 	<tr>
 		<td class="label"><label>{$PALANG.password}:</label></td>
-		<td><input class="flat" type="password" name="fPassword" /></td>
+		<td><input class="flat" type="password" name="fPassword" />{if $forgotten_password_reset}<br/>
+		<a href="password-recover.php">{$PALANG.pUsersLogin_password_recover}</a>{/if}</td>
```
cboltz: I prefer readable templates over cleanly formatted whitespace in the HTML output, so please don't hide the `if` in the middle of two lines.

upgrade.php
```diff
@@ -1633,6 +1633,20 @@ function upgrade_1835_mysql() {
...
+        if(!_mysql_field_exists($table, 'phone')) {
+            db_query_parsed("ALTER TABLE `$table` ADD COLUMN `phone` varchar(30) NOT NULL DEFAULT ''");
+        }
+        if(!_mysql_field_exists($table, 'email_other')) {
+            db_query_parsed("ALTER TABLE `$table` ADD COLUMN `email_other` varchar(255) NOT NULL DEFAULT ''");
+        }
+    }
+}
+
 # TODO MySQL:
```
cboltz: please use the _db_add_field() function instead of doing manual SQL queries. Also, please specify the encoding (unsing `{LATIN1}` or `{UTF-8}`) for all text fields you add. (See db_query_parsed() for a list of all suported `{...}` markers.)

functions.inc.php
DavidGoodwin: Why not use e.g. a truncation of a hash of the $CONF['setup_password'] - that'd give you something random to add into the hash creatio (and unique to this installation) to solve the problem?

users/password-change.php
```diff
@@ -0,0 +1,104 @@
...
+  $rel_path = './';
+  $context = 'admin';
+}
+require_once($rel_path . 'common.php');
+
+if ($context == 'admin' && !Config::read('forgotten_admin_password_reset') || $context == 'users' && !Config::read('forgotten_user_password_reset'))
+{
+    header('HTTP/1.0 403 Forbidden');
+    exit(0);
+}
```
cboltz: I'd prefer a die() with a useful message.
Attackers will know what a 403 means anyway, and normal users will have a hard time debugging an empty page.

users/password-change.php
```diff
@@ -0,0 +1,104 @@
...
+            flash_error($handler->errormsg);
+        } else {
+            $values = $handler->result;
+            $values[$handler->getId_field()] = $tUsername;
+            $values['password'] = $fPassword;
+            $values['password2'] = $fPassword2;
+            if ($handler->set($values) && $handler->store()) {
+                flash_info(Config::lang_f('pPassword_result_success', $tUsername));
+                header('Location: ' . dirname($_SERVER['REQUEST_URI']) . '/main.php');
+                exit(0);
```
cboltz: 'Location: main.php' (or 'Location: login.php' if you agree not to do the actual login here) will also work and makes the code much simpler.

users/password-recover.php
```diff
@@ -0,0 +1,124 @@
...
+  $rel_path = './';
+  $context = 'admin';
+}
+require_once($rel_path . 'common.php');
+
+if ($context == 'admin' && !Config::read('forgotten_admin_password_reset') || $context == 'users' && !Config::read('forgotten_user_password_reset'))
+{
+    header('HTTP/1.0 403 Forbidden');
+    exit(0);
+}
```
cboltz: As already noted in another file, please die() with a useful message.

users/password-recover.php
```diff
@@ -0,0 +1,124 @@
...
+        else
+        {
+            flash_error(Config::Lang('pPassword_recovery_no_alternative'));
+        }
+    }
+    else
+    {
+        flash_error(Config::Lang('pCreate_mailbox_username_text_error1'));
+    }
+}
```
cboltz: This file is leaking information in several ways:
- if a user exists
- if that user has a backup mail address and/or phone number defined

I know it sounds useless, but a message saying "We processed your request. If you entered a valid username, you'll receive a mail and/or SMS with a password reset code." would avoid that leak.

(We had a bugreport in the past about a similar information leak in login.php about "invalid username" vs. "wrong password" ;-)

users/password-change.php
```diff
@@ -0,0 +1,104 @@
...
+
+if ($context == 'admin' && !Config::read('forgotten_admin_password_reset') || $context == 'users' && !Config::read('forgotten_user_password_reset'))
+{
+    header('HTTP/1.0 403 Forbidden');
+    exit(0);
+}
+
+if ($_SERVER['REQUEST_METHOD'] == 'GET')
+{
+    $tUsername = safeget('username');
```
cboltz: I know we have both styles for historical reasons, but new code should follow https://sourceforge.net/p/postfixadmin/wiki/Coding_guidelines/ which means to have the opening curly brackets in the same line as the statement. (The coding guidelines contain a bit more than that, so please have a quick look at them.)

functions.inc.php
cboltz: That would be better than not including any secret at all, but
- the setup password might be empty (you can create your first admin with postfixadmin-cli, so you don't really need a setup password) - so don't rely on it
- re-using the same secret every time could allow to guess/brute force it if a user uses "lost password" several times
- this patch is security relevant, so it MUST be secure. Not "somewhat secure".

config.inc.php
```diff
@@ -562,6 +570,18 @@ function x_struct_admin_modify($struct) {
...
+$CONF['forgotten_user_password_reset'] = true;
+// Allows an admin to reset his forgotten password with a code sent by email/SMS
+$CONF['forgotten_admin_password_reset'] = true;
+
+// Clickatell gateway to send SMS code for password reset
+// API type: HTTP
+$CONF['clickatell_api_id'] = '';
+$CONF['clickatell_user'] = '';
+$CONF['clickatell_password'] = '';
+$CONF['clickatell_sender'] = '';
```
cboltz: I'd guess there are other services for sending SMS, so it might be a good idea to change this to `$CONF['sms_send_function'] = '';` and then provide the sendCodebySMS() function as an example directly in config.inc.php. Maybe even commented out - the api ID etc. need to be filled in anyway, so removing the comment signs wouldn't be hard.

config.inc.php
```diff
@@ -562,6 +570,18 @@ function x_struct_admin_modify($struct) {
 // http://www.php.net/manual/en/function.imap-open.php
 $CONF['create_mailbox_subdirs_hostoptions'] = array('');
 
+// Optional:
+// Allows a user to reset his forgotten password with a code sent by email/SMS
+$CONF['forgotten_user_password_reset'] = true;
+// Allows an admin to reset his forgotten password with a code sent by email/SMS
+$CONF['forgotten_admin_password_reset'] = true;
```
cboltz: I'd tend to have this defaulting to false ;-)

config.inc.php
gnanet: I also suggest to use some provider independent  solution, because I am currently investigating hand-in-hand with Clickatell delivery issues, where an SMS is reported as "delivered to device", but the SMS never arrives.

users/password-change.php
PowerKiKi: I was unaware of the fact that relative URL have been [officially part of the RFC](https://tools.ietf.org/html/rfc7231#section-7.1.2) for [over 5 years](https://trac.tools.ietf.org/wg/httpbis/trac/ticket/185). Thanks for letting me know.

users/password-change.php
```diff
@@ -0,0 +1,104 @@
...
+
+    if (empty($fPassword) or ($fPassword != $fPassword2)) {
+        $error = true;
+        flash_error(Config::lang('pPassword_password_text_error'));
+    } elseif (trim(strtoupper($tCode) != getPasswordRecoveryCode($tUsername))) {
+        flash_error(Config::lang('pPassword_code_text_error'));
+    } else {
+        session_regenerate_id();
+        $_SESSION['sessid']['username'] = $tUsername;
+        if ($context == 'users') {
```
PowerKiKi: If the login code can be refactored in a shared function, then I would vote in favor of auto-login after password reset.

I don't think it could be a security concern, since the user was just authorized to change his password anyway. And removing a step is indeed convenient for the user, because it's extremely likely that what he wanted to do in the first place was to login.

upgrade.php
cboltz: Thanks for the update.

Now if you look at _db_add_field(), you'll notice that it already calls _db_field_exists(), so you can also drop this check. (Yes, life can be simple ;-)

Oh, and please drop the _mysql suffix from the function name so that all database types get updated.

model/MailboxHandler.php
```diff
@@ -43,6 +43,11 @@ protected function initStruct() {
             # TODO: add virtual 'notified' column and allow to display who received a vacation response?
         );
 
+        if (Config::read('forgotten_user_password_reset')) {
+            $this->struct['phone'] = pacol(   1,          1,      0,      'text', 'pCreate_mailbox_phone'         , 'pCreate_mailbox_phone_desc'       , '');
+            $this->struct['email_other'] = pacol(   1,          1,      0,      'mail', 'pCreate_mailbox_email'         , 'pCreate_mailbox_email_desc'       , '');
+        }
+
```
cboltz: Much better, but the usual way to do this is to set display_in_list and display_in_form based on the config option (like $this->new is used for some fields). This way ensures the field gets filled with the correct default value if it's hidden from the user.

config.inc.php
```diff
@@ -119,6 +119,14 @@ function language_hook($PALANG, $language) {
...
 $CONF['admin_email'] = '';
 
+// Site admin name
+// This will be used as signature in notification messages
+$CONF['admin_name'] = 'Postmaster';
+
+// Site admin phone number
+// This will be used if a user cannot access his/her email and needs support
+$CONF['admin_phone'] = '';
+
```
cboltz: It seems you don't use $CONF['admin_phone'] anywhere, so please don't add it. We already have enough config options ;-)

model/MailboxHandler.php
stissot: This has been fixed: the additional fields for password reset are always added to the structure, but are displayed in the form only when the password reset feature is enabled.

config.inc.php
stissot: The SMS sending is now using a user-defined function, with an example for Clickatell provided in config file.

users/password-recover.php
stissot: The password recovery form is now displaying a generic "Your request has been processed, you should receive an email or SMS with instructions to reset your password" no matter if the account exist or has alternative contact info. In addition, the reply is delayed of ~2 sec to prevent brute force attacks.

users/password-recover.php
```diff
@@ -0,0 +1,124 @@
...
+    $fHeaders .= "Subject: " . encode_header(Config::Lang('pPassword_welcome')) . PHP_EOL;
+    $fHeaders .= "MIME-Version: 1.0" . PHP_EOL;
+    $fHeaders .= "Content-Type: text/plain; charset=utf-8" . PHP_EOL;
+    $fHeaders .= "Content-Transfer-Encoding: 8bit" . PHP_EOL . PHP_EOL;
+
+    $url = $_SERVER['REQUEST_SCHEME'] . '://' . $_SERVER['HTTP_HOST'] . dirname($_SERVER['REQUEST_URI']) . '/password-change.php?username=' . urlencode($username) . '&code=' . $code;
+    $fHeaders .= Config::lang_f('pPassword_recovery_email_body', $url);
+
+    return smtp_mail($to, Config::read('admin_email') , $fHeaders);
+}
```
stissot: Thanks, this has been fixed.

functions.inc.php
stissot: The implementation is now using the existing generate_password() function to generate a random 12 alphanum characters reset token unique to the user, which is stored in database, is valid for 1 hour and can only be used once.

config.inc.php
stissot: Indeed admin_phone was intended for letting the user know how to contact the postmaster if his account doesn't have alternative contact detail, but is not used anymore, I've removed it.

foobarable:
This feature is awesome. What is missing for this to be merged?

languages/en.lang
```diff
@@ -297,6 +309,7 @@ $PALANG['pAdminEdit_admin_result_success'] = 'The admin %s has been modified.';
 $PALANG['pUsersLogin_welcome'] = 'Mailbox users login to change your password and aliases.';
 $PALANG['pUsersLogin_username_incorrect'] = 'Your login is not correct. Make sure that you login with your email address!';
 $PALANG['pUsersLogin_password_incorrect'] = 'Your password is not correct!';
+$PALANG['pUsersLogin_password_recover'] = 'Did you forget your password?';
```
jdsn: clicking on a question feels a bit strange, why not putting it like this: `I forgot my password.` ?

IdleGandalf:
Any progress on this? It would be awesome to move the password resetting tasks to the user itself instead of the admin.

PowerKiKi:
I see that it now conflicts with latest dev. I am willing to resolve those, but only if @cboltz or @DavidGoodwin can confirm they still have interest in merging this ?

languages/en.lang
stissot: Fixed, the link is now "I forgot my password". Thanks for the suggestion

stissot:
I just rebased the PR from master, can @cboltz or @DavidGoodwin merge ?

DavidGoodwin:
Thanks for the 'poke'....

cboltz:
It seems the rebase killed some of the changes, and we ended up with the insecure version that sends out a guessable hash (based on username + date) :-(  Also, only the mysql database layout gets updated and some other minor things mentioned above also should get fixed.

Please review the comments in this pull request again and send a new pull request that fixes these issues.

I disabled $CONF['forgotten_user_password_reset'] and $CONF['forgotten_admin_password_reset'] in config.inc.php for now and added a warning. Feel free to re-enable it with your pull request to make the password reset secure.

cboltz:
Any news on this?

I'm not willing to carry the insecure code (even if disabled by default in the config), so I'm afraid I'll have to completely revert this pull request if nobody fixes the security issues within the next two weeks.

PowerKiKi:
@cboltz I'll see what I can do about that...

PowerKiKi:
@cboltz, please take a look at #79, it should have everything this PR used to have.

PR #23, opened 2017-02-03
Fix security hole in AliasHandler

Janfred:
Without this fix it is possible to delete a protected alias via editing
the request parameter of the alias to delete.

A simple example: If the alias abuse@example.com is defined as default alias and the alias other@example.com exists, you can just change the parameter in the delete url:

The delete url for other@example.com will probably look like this:
`https://pfadmin.example.com/delete.php?table=alias&delete=other%40example.com&token=<CSRF-Token>`

You can just change the request to 
`https://pfadmin.example.com/delete.php?table=alias&delete=abuse%40example.com&token=<CSRF-Token>`
and Postfixadmin will delete the default alias.

The suggested fix just adds a condition to the delete function in AliasHandler which will fail if `$this->can_delete` is false.

### DIFF
--- a/model/AliasHandler.php
+++ b/model/AliasHandler.php
@@ -441,6 +441,11 @@ class AliasHandler extends PFAHandler {
             return false;
         }
 
+        if ($this->can_delete) {
+            $this->errormsg[] = Config::Lang_f('protected_alias_cant_be_deleted', $this->id);
+            return false;
+        }
+
         db_delete('alias', 'address', $this->id);
 
         list(/*NULL*/,$domain) = explode('@', $this->id);
--- a/model/AliasHandler.php
+++ b/model/AliasHandler.php
@@ -441,7 +441,7 @@ class AliasHandler extends PFAHandler {
             return false;
         }
 
-        if ($this->can_delete) {
+        if (!$this->can_delete) {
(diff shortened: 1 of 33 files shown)


model/AliasHandler.php
```diff
@@ -441,6 +441,11 @@ public function delete() {
             return false;
         }
 
+        if ($this->can_delete) {
```
cboltz: Nice catch! Thanks for reporting it!

However I wonder if this condition is the wrong way round - I'd think we need
`    if ( ! $this->can_delete) {`
(note the added '!')

Janfred: Oooops...
I've forgotten the `! ` while transferring the patch from my production system to my local git copy...

cboltz:
This is CVE-2017-5930.

cboltz:
I released PostfixAdmin 3.0.2 which includes this fix and some non-security bugfixes.

Thanks again for reporting this and for submitting the fix.

xpunkt:
would be nice to see 3.0.2 github tagged :)

but thumps up that it is on github, i would try to make a gentoo ebuild to get cutting edge

DavidGoodwin:
@xpunkt https://github.com/postfixadmin/postfixadmin/releases/tag/postfixadmin-3.0.2 good enough?

PR #24, opened 2017-02-11
Broadcast improvements

Janfred:
This pull request improves the broadcast message feature.

The reason I've submitted this pull request is, that I've had a use case where I
wanted to notify all Mailbox users. (In this special case because of an update
of the web mailer)  
Since this information is useless for forwards who don't deliver to local
mailboxes I've increased the spam for all of these users by sending a mail to
their forward.

I've also found that it might be desirable that you can give local admins the
right to send broadcast messages to "their" users.
This feature can be enabled by setting the config option
`sendmail_only_global_admin` to `NO`. I've set the default to `YES` to have not
enabled it accidentally if upgrading postfixadmin

Since some of the information might also be just for one domain (e.g. "We'll
deploy DKIM on this domain" or "We'll change the MX servers for this domain")
I've implemented that you can choose the domains you want to send the broadcast
mail to.


broadcast-message.php
```diff
@@ -27,14 +27,19 @@
 
 require_once('common.php');
 
-authentication_require_role('global-admin');
+if ($CONF['sendmail_only_global_admin'] == 'YES' ) {
```
DavidGoodwin: so, if you're upgrading and haven't added sendmail_only_global_admin to your config then any admin can then use this? 

( i think you need to flip the logic and have a config lg for granting permission to normal admins, so by default it remains a global admin only feature)

Janfred: I can change that, but IIRC the config.inc.php should also be subject of upgrading since all defaults are defined there.

cboltz: That's the theory, but you'll never know about practise ;-)  (if someone edited config.inc.php instead of creating a config.local.php, we might end up with a config.inc.php.rpmnew). I also agree with David that you should flip the logic - "disable_feature = yes" is terribly confusing, "enable_feature = no" is much better.

BTW: Instead of checking $CONF directly, please use `Config::bool('broadcast_domain_admins')`. I know we didn't migrate everything yet, but new code should use it ;-)

Janfred: I've changed that.

Speaking of not migrated files: The indentation of the file looks very scary...

broadcast-message.php
```diff
@@ -47,23 +52,38 @@
...
-      {
+      $table_alias = table_by_key('alias');
+
+      $recipients = [];
+
+      foreach( $wanted_domains as $domain) {
+         $q = "SELECT username from $table_mailbox WHERE domain = '$domain'";
+         if(!isset($_POST['mailboxes_only']) || $_POST['mailboxes_only'] != "on") {
+            $q .= " UNION SELECT GOTO FROM $table_alias WHERE domain = '$domain' " .
+                  " AND goto NOT IN ($q)";
```
cboltz: Sorry for the delay, I'm too busy nowadays :-(

Instead of running multiple queries in a loop, please use an `IN` clause to match all requested domains in one query - something like `WHERE domain IN ('foo', 'bar')`

To make your life easier - you can use `db_in_clause()` to generate this part of the query ;-)

On the long term, using MailboxHandler and AliasHandler would be even better - but you don't need to do this unless you really want ;-)

Janfred: No problem, I've been busy too...

I've changed the `foreach`-Loop to `db_in_clause` and it seems to work (I've just tested it once on my server...)

The refactoring to move this code to the Mailbox-/AliasHandler is a to complicated job to do it during exams, maybe I'll have a look at this when I'm less busy then now

broadcast-message.php
```diff
@@ -47,23 +52,35 @@
...
-
-      $result = db_query ($q);
-      if ($result['rows'] > 0)
-      {
+      $table_alias = table_by_key('alias');
+
+      $recipients = [];
+
+      $q = "SELECT username from $table_mailbox WHERE ".db_in_clause("domain", $wanted_domains);
+      if(!isset($_POST['bailboxes_only']) || $_POST['mailboxes_only'] != "on") {
```
BigMichi1: shouldn't this be 'mailboxes_only' instead of 'bailboxes_only'

Janfred: It should.
This is what I meant by "just tested it once on my server"....

languages/de.lang
```diff
@@ -331,6 +331,8 @@ $PALANG['pBroadcast_name'] = 'Ihr Name';
 $PALANG['pBroadcast_success'] = 'Ihre Rundmail wurde gesendet.';
 $PALANG['pAdminMenu_broadcast_message'] = 'Rundmail';
 $PALANG['pBroadcast_error_empty'] = 'Die Felder Name, Betreff und Nachricht dürfen nicht leer sein!';
+$PALANG['broadcst_mailboxes_only'] = "Nur an Postfächer senden";
+$PALANG['broadcst_to_domains'] = "Senden an Domains:";
```
cboltz: nice typo - please change broadcst_* to broadcast_* ;-) (hint: running language-update.sh helps to find such issues)

templates/broadcast-message.tpl
```diff
@@ -22,6 +22,18 @@
 		<td><textarea class="flat" cols="40" rows="6" name="message"></textarea></td>
 	</tr>
 	<tr>
+		<td class="label"><label>{$PALANG.broadcast_mailboxes_only}</label></td>
+		<td><input type="checkbox" name="mailboxes_only"/></td>
```
cboltz: Your PHP code expects the checkbox value to be "on", but I wouldn't rely on that ;-)

It's better to use ` type="checkbox" value='1'`

broadcast-message.php
```diff
@@ -49,23 +54,35 @@
...
-
-      $result = db_query ($q);
-      if ($result['rows'] > 0)
-      {
+      $table_alias = table_by_key('alias');
+
+      $recipients = [];
+
+      $q = "SELECT username from $table_mailbox WHERE ".db_in_clause("domain", $wanted_domains);
+      if(!isset($_POST['mailboxes_only']) || $_POST['mailboxes_only'] != "on") {
```
cboltz: With the checkbox value changed to "1", you can simplify this condition:

        if (intval(safepost('mailboxes_only')) == 0) {

BTW: a space after `if` would be nice.

Janfred: Done. Sorry for this, I usually program in rails and don't have to mess with plain html.

languages/fr.lang
```diff
@@ -324,7 +324,9 @@ $PALANG['pBroadcast_title'] = 'Envoyer un message général';
 $PALANG['pBroadcast_name'] = 'Votre nom';
 $PALANG['pBroadcast_success'] = 'Votre message général a été envoyé.';
 $PALANG['pAdminMenu_broadcast_message'] = 'message général';
-$PALANG['pBroadcast_error_empty'] = 'Les champs "Nom", "Sujet" et "Message" ne peuvent pas être vides !';
+$PALANG['pBroadcast_error_empty'] = 'Les champs Nom, Sujet et Message ne peuvent pas être vides!';
```
Janfred: Done. ;)

languages/de.lang
Janfred: sorry. fixed.

cboltz:
Thanks for the work on this and the quick updates. This feature is finally ready for a merge :-)

PR #25, opened 2017-02-12
Added support for mkpasswd encryption method

Sopsy:
This makes it possible to use postfixadmin with eg. SHA-512 and with a rounds -setting on a server without Dovecot installed.

Useful when installed on a different server than where Dovecot resides, eg. a webhost.

FYI: mkpasswd is installed on debian/ubuntu with the package "whois":

New config keys:
$CONF['mkpasswd'] = "/usr/bin/mkpasswd";
$CONF['mkpasswd_rounds'] = 30000; // Only supported by sha-xxx

New values for encrypt:
$CONF['encrypt'] = 'mkpasswd:des'; // Might not be supported by dovecot/postfix at all
$CONF['encrypt'] = 'mkpasswd:md5';
$CONF['encrypt'] = 'mkpasswd:sha-256';
$CONF['encrypt'] = 'mkpasswd:sha-512';

### DIFF
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -1007,6 +1007,85 @@ function pacrypt ($pw, $pw_db="") {
         }
     }
 
+    elseif (preg_match("/^mkpasswd:/", $CONF['encrypt'])) {
+        $split_method = preg_split ('/:/', $CONF['encrypt']);
+        $method       = strtolower($split_method[1]);
+
+        if (!empty($pw_db)) {
+            // If we are verifying the password, try to figure out the old method
+            if (substr($pw_db, 0, 3) == '$1$') {
+                $method = 'md5';
+            } elseif (substr($pw_db, 0, 3) == '$5$') {
+                $method = 'sha-256';
+            } elseif (substr($pw_db, 0, 3) == '$6$') {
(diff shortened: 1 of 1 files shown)


cboltz:
It seems [crypt()](http://php.net/crypt) also supports these hashing methods and specifying the number of rounds. Would that be an alternative?
(I'd prefer it because it doesn't rely on an external program.)

Note: The 'system' hashing type uses crypt(), but it doesn't allow to specify the hashing method yet. Adding support for that (as "system:sha-256:30000") would be nice ;-)
Actually there's already a patch that promises to do something like that - https://sourceforge.net/p/postfixadmin/patches/121/ (untested).

Sopsys:
I'll check if I can rewrite the code to use crypt() instead of mkpasswd. That just seemed to be the easy way out. I'll return to this later.

Thanks for the idea!

cboltz:
You might want to look at #32 ;-)  I only had a quick look at it, but it probably does what you want.

DavidGoodwin:
Closing - I think this is handled in the code in #32 and there's no benefit to this patch being merged (aside from making the hashing options yet more complex)

PR #26, opened 2017-02-14
added config option to disable "edit_alias" function for users

medarion:
This commit adds the possibility to prevent users from setting aliases.

### DIFF
--- a/users/edit-alias.php
+++ b/users/edit-alias.php
@@ -22,14 +22,14 @@ $rel_path = '../';
 require_once('../common.php');
 $smarty->assign ('smarty_template', 'users_edit-alias');
 
-// is edit-alias support enabled in $CONF ?
-if($CONF['edit_alias'] == 'NO') {
-  header ("Location: $Return_url");
-  exit(0);
-}
-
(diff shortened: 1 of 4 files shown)


users/edit-alias.php
```diff
@@ -22,6 +22,12 @@
 require_once('../common.php');
 $smarty->assign ('smarty_template', 'users_edit-alias');
 
+// is edit-alias support enabled in $CONF ?
+if($CONF['edit_alias'] == 'NO') {
```
cboltz: Please change this condition to
`if (! Config::bool('edit_alias')) {`

Also, please move this section after `$USERID_USERNAME = ...` so that the authentification check stays the first thing edit-alias does.

medarion:
Thanks for the comments. I applied the changes and also fixed the wrong redirect to $RedirectUrl, which wasn't defined.

cboltz:
Thanks for the update, I'll merge it.

Minor nitpicking: you chose a strange place for `$CONF['edit_alias']` in config.inc.php. I'll move it next to `$CONF['alias_control']` after merging.

PR #31, opened 2017-03-06
Adding BASIC-Auth-patch

dennisse:
From https://sourceforge.net/p/postfixadmin/patches/131/

### DIFF
--- a/setup.php
+++ b/setup.php
@@ -318,7 +318,11 @@ else
     print "<p>Everything seems fine... attempting to create/update database structure</p>\n";
     require_once($incpath.'/upgrade.php');
 
-    $tUsername = '';
+    if ( isset($_SERVER['REMOTE_USER']) ) {
+        $tUsername = strtolower($_SERVER['REMOTE_USER']);
+    } else {
+        $tUsername = '';
+    }
     $setupMessage = '';
     $lostpw_error = 0;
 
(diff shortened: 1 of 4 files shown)


cboltz:
I'm not against supporting basic auth, but I don't like the idea to use it whenever $_SERVER['REMOTE_USER'] is provided. This might even be a security risk if someone is able to place a .htaccess file on the domain (maybe in another subdirectory) where PostfixAdmin is running.

There should be a config option to enable basic auth (disabled by default) - and if enabled, it should probably *replace* the password-based auth.  Ideally this config option should allow to specify which $_SERVER variable to use to allow support for other HTTP headers, maybe

    $_CONF['http_auth'] = "";  # disabled
    $_CONF['http_auth'] = "REMOTE_USER';  # use $_SERVER['REMOTE_USER'] 

Enabling this also means to hide the then-superfluous password field on the login page, and to make the username field pre-filled (not editable) with the username. In other words: login.php would *display* the username, and the only thing a user can do is to choose another language and then click the login button.

I'd even consider to hide the password fields when creating/editing an admin - if the authentification is done externally, there's no need to set admin passwords.

Mailbox passwords are needed for POP3, IMAP and SMTP auth, so don't hide the password fields for them ;-)

setup.php
```diff
@@ -351,14 +355,26 @@
...
+                $values = array(
+                    'username'      => strtolower($_SERVER['REMOTE_USER']),
+                    'password'      => safepost('password'),
+                    'password2'     => safepost('password2'),
+                    'superadmin'    => 1,
+                    'domains'       => array(),
+                    'active'        => 1,
+                );
+            } else {
+                $values = array(
```
jdsn: looks like duplication, I'd rather set all common values first and then add the differing value in an if clause

DavidGoodwin:
https://www.php.net/manual/en/features.http-auth.php

I don't think this patch should be merged; but once again some sort of hook behaviour would be the best approach to handle this (see how e.g. wordpress handles this, which gives an idea of how it could be done with some sort of hook .... https://core.trac.wordpress.org/browser/tags/5.3/src/wp-includes/pluggable.php#L521 )

PR #32, opened 2017-03-07
multiple hash algorithms using system's crypt in pacrypt

snuggeman:
This patch adds another $CONF['encrypt'] type, 'php_crypt', which uses the PHP crypt() function, which in turn uses the system's crypt() (much like the current 'system' type in postfixadmin-2.3.6)

The difference to the current 'system' type is that you can choose which hash type you want to use for new passwords. (currently DES, MD5, Blowfish, SHA256, SHA512; defaults to MD5)

Example:
$CONF['encrypt'] = 'php_crypt:SHA512';

The current 'system' type only generates MD5 hashes with 2-character salt.

I did not have 'dovecotpw' on my system so that was not an option.

It does not interfere with existing 'encrypt' types, but could be incorporated into the 'system' type.
Tested on CentOS 5, 6 and 7, Dovecot 2 and Postfix.

### DIFF
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -1013,12 +1013,96 @@ function pacrypt ($pw, $pw_db="") {
             $password = rtrim($password);
         }
     }
+	
+    elseif (substr($CONF['encrypt'], 0, 9) === 'php_crypt') {
+        // use PHPs crypt(), which uses the system's crypt()
+        // same algorithms as used in /etc/shadow
+        // you can have mixed hash types in the database for authentication, changed passwords get specified hash type
+        // the algorithm for a new hash is chosen by feeding a salt with correct magic to crypt()
+        // set $CONF['encrypt'] to 'php_crypt' to use the default MD5 crypt method
+        // set $CONF['encrypt'] to 'php_crypt:METHOD' to use another method; methods supported: DES, MD5, BLOWFISH, SHA256, SHA512
+        // tested on linux
+
+        if (strlen($pw_db) > 0) {
+            // existing pw provided. send entire password hash as salt for crypt() to figure out
+            $salt = $pw_db;
+        } else {
+            // no pw provided. create new password hash
+            if(strpos($CONF['encrypt'], ':') !== false) {
+				// use specified hash method
+                $split_method = preg_split ('/:/', $CONF['encrypt']);
+                $salt_method = $split_method[1];
+            }
+            // create appropriate salt for selected hash method
+            $salt = generate_crypt_salt($salt_method);
+        }
+        // send it to PHPs crypt()
+        $password = crypt($pw, $salt);
+    }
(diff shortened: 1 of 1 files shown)


functions.inc.php
```diff
@@ -1013,6 +1013,32 @@ function pacrypt ($pw, $pw_db="") {
...
+        // tested on linux
+
+        if (strlen($pw_db) > 0) {
+            // existing pw provided. send entire password hash as salt for crypt() to figure out
+            $salt = $pw_db;
+        } else {
+            // no pw provided. create new password hash
+            if(strpos($CONF['encrypt'], ':') !== false) {
+				// use specified hash method
+                $split_method = preg_split ('/:/', $CONF['encrypt']);
```
cboltz: `explode(':', $CONF['encrypt'])` is probably faster ;-)

BTW: The comment line above has tabs instead of spaces.

grubersh:
@cboltz We have the same issue with only MD5 hashes being generated. As we are using php-fpm with chroot postfixadmin has no access to doveadm. So it would be great if this patch could be include into postfixadmin so that other schemes could be configured!

snuggeman:
patch for 3.1, see: https://github.com/postfixadmin/postfixadmin/pull/52

functions.inc.php
```diff
@@ -1021,6 +1047,64 @@ function pacrypt ($pw, $pw_db="") {
     return $password;
 }
 
+// used for php_crypt method
+function random_string($characters, $length) {
+    $string = '';
+    for ($p = 0; $p < $length; $p++) {
+        // should really use a stronger randomness source
+        $string .= $characters[mt_rand(0, strlen($characters)-1)];
```
CyberFoundry: mt_rand is NOT cryptographically secure and should not be used for salt generation. Check out random_bytes /  openssl_random_pseudo_bytes /  mcrypt_create_iv functions.

functions.inc.php
```diff
@@ -1013,6 +1013,32 @@ function pacrypt ($pw, $pw_db="") {
...
+            if(strpos($CONF['encrypt'], ':') !== false) {
+				// use specified hash method
+                $split_method = preg_split ('/:/', $CONF['encrypt']);
+                $salt_method = $split_method[1];
+            }
+            // create appropriate salt for selected hash method
+            $salt = generate_crypt_salt($salt_method);
+        }
+        // send it to PHPs crypt()
+        $password = crypt($pw, $salt);
```
CyberFoundry: Adding BASE64 password encoding would be a nice enhancement.

CyberFoundry: And also password scheme identification. E.g. {SSHA.HEX}, {SHA512-CRYPT.B64} ...

snuggeman:
see also #170

PR #33, opened 2017-03-11
Update French language

acs-ferreira:


cboltz:
Thanks for the translation update, it looks mostly good (says someone who doesn't understand French ;-)

I noticed that you added a space in front of several, but not all exclamation marks. This would be an error in English and German, so - is French different here?

Also, please remove the `# XXX` marker in lines you translated to make it easier for the next one who looks for missing translations ;-)

Oh, and if you want, you can add yourself in the file header ;-)

acs-ferreira:
Yes, in French add a space between the end of line and exclamation mark.
Update the file accordingly.

cboltz:
Thanks for the quick update - and I even learned something about spacing around exclamation marks in French :-)

PR #34, opened 2017-03-12
Custom mysqli port

AnthoDingo:
Add custom mysqli port in connection string

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -86,6 +86,8 @@ $CONF['database_host'] = 'localhost';
 $CONF['database_user'] = 'postfix';
 $CONF['database_password'] = 'postfixadmin';
 $CONF['database_name'] = 'postfix';
+// Default mysql port for mysqli connection
+$CONF['database_port'] = '3308';
 // If you need to specify a different port for a MYSQL database connection, use e.g.
 //   $CONF['database_host'] = '172.30.33.66:3308';
 // If you need to specify a different port for POSTGRESQL database connection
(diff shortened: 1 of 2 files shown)


config.inc.php
```diff
@@ -86,6 +86,8 @@ function language_hook($PALANG, $language) {
 $CONF['database_user'] = 'postfix';
 $CONF['database_password'] = 'postfixadmin';
 $CONF['database_name'] = 'postfix';
+// Default mysql port for mysqli connection
+$CONF['database_port'] = '3308';
```
cboltz: PostgreSQL also looks at $CONF[database_port], and from looking at the code in functions.inc.php, I'm afraid this change would cause breakage for all PostgreSQL users because PostfixAdmin would always try to connect to the MySQL port ;-)

I'm a bit undecided what the best way to handle this is - possible options include
* adding a $CONF[database_port_mysql] (argh, another config option)
* having $CONF[database_port] = '0' as default, which would mean to use the default port of each database

BTW: For mysql, something like $CONF[database_host] = '127.0.0.1:3308' worked. Does this also work for mysqli?

AnthoDingo: URI host:port not work for mysqli. :(
Mysqli constructor : http://php.net/manual/fr/mysqli.construct.php

DavidGoodwin:
https://github.com/postfixadmin/postfixadmin/commit/4b999b3f6bb78f595af2a7294f6b231de8f61fc4 is relevant

config.inc.php
DavidGoodwin: * leave the port undefined by default.
 * pass in the default value to the appropriate DB driver if it's not defined in the config. ?

DavidGoodwin:
closing; https://github.com/postfixadmin/postfixadmin/commit/4b999b3f6bb78f595af2a7294f6b231de8f61fc4 should fix this.

PR #36, opened 2017-03-17
Add a .gitignore file which follows the svn:ignore

isleshocky77:
It appears that pull requests are coming in from github via git now. This PR is useful for anyone managing their postfixadmin code through git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+/templates_c/*.tpl.php
+/templates_c/*menu.conf.config.php
+


cboltz:
Thanks!

Looking at the files in templates_c/, it turns out that `*menu.conf.config.php` is wrong (svn:ignore was also wrong, so not your fault) and should be `*.menu.conf.php`. I'll fix this after merging your pull request ;-)

PR #37, opened 2017-03-17
Move config.inc.php to config.inc.php.dist for security purposes

isleshocky77:
Force the creation of a config.inc.php file which is ignored by VCSs and not accidentally overwritten by upgrades.


cboltz:
Did you ever read about using a config.local.php file? ;-)  Just create config.local.php and add your settings there. It will automatically be included at the end of config.inc.php, and it will never be overwritten by package managers or git.

Therefore renaming config.inc.php to something that is a) readable to everybody (not "protected" by PHP) and b) not loaded by default looks like a bad idea.

BTW: I don't see why the change you propose would be "for security purposes" - can you please explain that?

isleshocky77:
@cboltz Nope. Sorry about that. I'll kill this.

PR #43, opened 2017-04-10
Fix for PHP 7+

tswestendorp:
PHP 7+ produces a warning when assigning an empty string to a string offset


cboltz:
Your fix is technically correct, but grep just told me that this variable isn't used anymore. Can you please change your pull request to simply remove it? ;-)

BTW: the long-term goal is to get rid of variables.inc.php, and I wouldn't be surprised if it contains more variables that are no longer used.

tswestendorp:
> Can you please change your pull request to simply remove it? ;-)
done

PR #49, opened 2017-06-14
 Updated version of virtual vacation script with Email::Sender

gianlucagiacometti:
Mail::Server  ==> Email::Server
Main update.
MAIL::Server is deprecated and the script neeeded the new library.

GetOpt::Std ==> GetOpt::Long
On my Ubuntu 16.04 servers with Perl 5.22 I had problems to install GetOpt::Std, cpan request was to force update to Perl 5.26, which I didn't want to.
I changed the usage syntax, which now is: vacation.pl [-t] -f {sender} -d {recipient}

Logger initialization
I changed the logger initialization and put it into the config section

Version update
I choosed to start a version 5.0 of the script since the updates are rather deep.

### DIFF
--- a/VIRTUAL_VACATION/INSTALL.TXT
+++ b/VIRTUAL_VACATION/INSTALL.TXT
@@ -90,7 +90,7 @@ Define the transport type in the Postfix master file:
 
 #/etc/postfix/master.cf:
 vacation    unix  -       n       n       -       -       pipe
-  flags=Rq user=vacation argv=/usr/lib/postfixadmin/vacation.pl -f ${sender} -- ${recipient}
+  flags=Rq user=vacation argv=/usr/lib/postfixadmin/vacation.pl -f ${sender} -d ${recipient}
 
 5. Setup the transport maps file
 --------------------------------
--- a/VIRTUAL_VACATION/vacation.pl
+++ b/VIRTUAL_VACATION/vacation.pl
@@ -1,4 +1,39 @@
-#!/usr/bin/perl -X
+#!/usr/bin/perl
+
+# Virtual Vacation 5.0
+#
+# $Revision: 1 $
+#
+# Updates:
+# 2017/06/13  Gianluca Giacometti <php at gianlucagiacometti dot it>
+#             Updated the usage syntax implementing GetOpt::Long
+#             Removed deprecated Mail::Sender and implemented Email::Sender
+#
+
+#
+# >> NEW <<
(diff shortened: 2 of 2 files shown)


DavidGoodwin:
Thanks for submitting this. I'll try and look at it soon (and try and run it myself).

cboltz:
I only had a quick look at the changes, but have some concerns:

* is the command line syntax change really needed? There should be a *very good* reason for forcing all our users to edit the postfix config (having some trouble with GetOpt::Std on your system doesn't sound like a good reason - it's a widely used module and I'm quite sure it works for nearly everybody, because otherwise we would have received lots of complaints. AFAIK it's even part of the perl base package, so I'm surprised you need to install it from cpan.)
* you replaced some $log* config options with hardcoded values
* "Uncomment as needed" sounds scary because it means users have to edit the script (not only the config file)

Switching to Email::Sender makes sense, but I don't like the changes listed above.

@DavidGoodwin: We should create a 3.0 branch (in case we want/need to ship another 3.0.x release) before changing the perl requirements of vacation.pl. I'm not sure if the branch should be based on current trunk or on trunk from some weeks ago - the mysqli fixes should go into 3.0.x, but I'm not sure about the new features.

gianlucagiacometti:
@cboltz 
I totally agree with you.
Please consider that I have no confidence at all with perl and all your concerns are related to my lack of knowledge.

- I was not able with Getopt::Long to reproduce the old usage syntax. Anyway, I checked better and realised that it's Getopt::Std and not GetOpt::Std, so we can go back to the old way avoiding syntax changes and complaints. I would suggest anyway a syntax change in your new release because it seems more standard without that 'strange' -- (if we can say so).

- I unsuccessfully tried to pass variables to my $conf = q(. I agree that would be better to separate code and config. So I moved that code as is under the config section.

- Same thing for the parameters of the transport. It would be much better to prepare a hash and pass it to the transport, being careful that, if I got well, Email::Sender::Transport::SMTPS is a little bit picky with such parameters (i.e. if ssl != '' then all parameters related to authentications are required and expected). So it should be well stated in the config comments 'you should know what you're doing', hoping Email::Sender documentation to become more clear and deep in future.

Thanks again to you both

DavidGoodwin:
i'm inclined to accept/merge #61 instead ....

DavidGoodwin:
#61 was merged ... I'm closing this as it's no longer needed. 

Thanks for providing the pull request.

PR #50, opened 2017-06-23
Update broadcast-message.php to send broadcast message to active users only

danmdm:
Send broadcast message to active users only. 
Without "active=1" condition in WHERE will clause a lot of "Recipient address rejected: User unknown in virtual mailbox table."

### DIFF
--- a/broadcast-message.php
+++ b/broadcast-message.php
@@ -61,9 +61,9 @@ if ($_SERVER['REQUEST_METHOD'] == "POST")
 
       $recipients = [];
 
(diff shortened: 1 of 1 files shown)


DavidGoodwin:
That works fine for MySQL - but for PostgreSQL the query needs changing to use the db_bool() function (I think it is).

cboltz:
It's `db_get_boolean(true)` ;-)

BTW: You don't need to add `active` to the SELECT part of the query - just using it in the WHERE section is fine.

cboltz:
Merged, I'll do the fixes mentioned above myself ;-)

PR #52, opened 2017-06-25
multiple hash algorithms using crypt (pfa 3.1)

snuggeman:
(for postfixadmin 3.1)

This patch adds another $CONF['encrypt'] type, 'php_crypt', which uses the PHP crypt() function, which in turn uses the system's crypt() (much like the current 'system' type in postfixadmin)

The difference to the current 'system' type is that you can choose which hash type you want to use for new passwords. (currently DES, MD5, Blowfish, SHA256, SHA512; defaults to MD5)

Example:
$CONF['encrypt'] = 'php_crypt:SHA512';

The current 'system' type only generates MD5 hashes with 2-character salt.

I did not have 'dovecotpw' on my system so that was not an option.

It does not interfere with existing 'encrypt' types, but could be incorporated into the 'system' type.
Tested on CentOS 5, 6 and 7, Dovecot 2 and Postfix.


DavidGoodwin:
out of stupidity, how does e.g. dovecot or postfix cope with you using e.g. a crypt() password or otherwise. 

Do you have examples of appropriate config bits for them?

snuggeman:
My postfix uses dovecot for auth:

# authentication
smtpd_sasl_auth_enable = yes
smtpd_sasl_type = dovecot
smtpd_sasl_path = /var/run/dovecot/auth-client
smtpd_sasl_security_options = noanonymous
broken_sasl_auth_clients = yes


and in my Dovecot config I have


in 10-auth.conf:
!include auth-sql.conf.ext


in auth-sql.conf.ext:
passdb {
   driver = sql
   args = /etc/dovecot/sql.conf.ext
}
userdb {
   driver = sql
   args = /etc/dovecot/sql.conf.ext
}

in sql.conf.ext:
default_pass_scheme = CRYPT



I suppose this last bit contains the magic!


/m

Den 2017-07-21 kl. 15:45, skrev David Goodwin:

snuggeman:
newer version: #170

DavidGoodwin:
#170 is merged.

PR #55, opened 2017-07-05
Fixed: php 5.4 array shortcut syntax errors on 5.3

dresken:


DavidGoodwin:
I'm inclined to think we should stop supporting any PHP version below 5.6 ( http://php.net/eol.php ). But I know some LTS releases will have old variants of PHP installed.

PR #60, opened 2017-07-17
Security fixes

Vilican:
I have fixed 2 non-critical security vulnerabilities:

 * login forms don't have CSRF tokens
 * links using target="_blank" in footer don't use rel="noopener"

### DIFF
--- a/login.php
+++ b/login.php
@@ -27,7 +27,6 @@
  *  lang
  */
 
-//define('POSTFIXADMIN_LOGOUT', 1);
 require_once('common.php');
 
 if($CONF['configured'] !== true) {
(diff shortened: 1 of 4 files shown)


login.php
```diff
@@ -23,10 +23,11 @@
  *
  *  fUsername
  *  fPassword
+ *  token
  *  lang
  */
 
-define('POSTFIXADMIN_LOGOUT', 1);
+//define('POSTFIXADMIN_LOGOUT', 1);
```
cboltz: Most of your changes make sense and are a good idea, but - why do you drop the POSTFIXADMIN_LOGOUT define? It is used in common.php to log out a user whenever going to login.php, and without that define that won't happen (and you are still logged in)...

Vilican: If you look carefuly on the code in common.php, you will see this:

```
if (defined('POSTFIXADMIN_LOGOUT')) {
    session_unset();
    session_destroy();
    session_start();
}
```

My code stores CSRF token in `$_SESSION` and this code deletes the session (including the token).

If you actually wanted to destroy session (log out) of the old user whenever new user logs in, you would use `session_regenerate_id (true)` - the true means destroy old session.

cboltz: Feel free to adjust the logout code with your proposal ;-)

I just tested your change and always get "invalid token" when trying to login :-( Can you please check what is wrong?

Vilican:
I have added back the logout mechanism. Also, I forgot to commint the change in template with CSRF token - fixed now.

PR #62, opened 2017-07-19
FetchmailHandler: Update to let the user choose from mailboxes and aliases

jcmondragon:
- Create an instance to AliasHandler in order to get the list of active
  aliases so they can be merged with the list of mailboxes.

### DIFF
--- a/model/FetchmailHandler.php
+++ b/model/FetchmailHandler.php
@@ -47,9 +47,12 @@ class FetchmailHandler extends PFAHandler {
         );
 
         # get list of mailboxes (for currently logged in user)
-        $handler = new MailboxHandler(0, $this->admin_username);
-        $handler->getList('1=1');
-        $this->struct['mailbox']['options'] = array_keys($handler->result);
+        # Added instance to show list of active alias (for currently logged in user)
+        $handler_mailbox = new MailboxHandler(0, $this->admin_username);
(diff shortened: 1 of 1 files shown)


DavidGoodwin:
I've never used the fetchmail functionality, so I don't really understand the point of this change. Why would you want the list to contain aliases as well as mailboxes?

jcmondragon:
Thank you for considering.
When the mail server is strictly internal and you receive mail from internet through a hosted mail service like Dreamhost, you use fetchmail to retrieve that external mail. If you need that some selected internal users can receive external mail strictly through an alias (not directly). You can create a hosted mailbox (in Dreamhost) which would be fetched from a local alias so the local alias deliver the mail to those selected users, but native source selection in the postfixadmin fetchmail module only shows mailboxes, I created an instance to AliasHandler in order to show local aliases too.
Please let me know if you have any other comment.
BTW- works nice!!
Regards.

cboltz:
I'm also not a big fan of this change because the typical usecase for fetchmail is to get it into a local mailbox.

Also offering aliases as possible target is not the end of the world ;-) but IMHO it's not really needed because in the end these aliases point to a mailbox. OTOH, it will make the selection list much longer, without benefit for average users.

For your usecase - would it be an option to create a mailbox, change it to forward-only and change this mailbox' alias to the target(s) you need?
This would give you what you want/need, while not flooding the selection list with all aliases.

PR #64, opened 2017-07-23
[WIP] add Dockerfile

J0WI:


J0WI:
Would love to see Postfixadmin in the Dockers standard library aka "Official Images".

There are also some administrative steps to fulfill:
https://github.com/docker-library/official-images#contributing-to-the-standard-library

DavidGoodwin:
Hi - thanks for taking the time to contribute this. I suppose this is another thing I should add to my 'todo' list.....

J0WI:
@volker-raschek I added my WIP postgresql container, but it is currently not working for me. Maybe you can help me finding the missing dependencies.

`setup.php` claims that `php5-pgsql` is missing, but this package is not available for PHP 7. Anyway, the correct way to install PHP extensions in Docker is over `docker-php-ext-install`.

J0WI:
I found the issue in my Dockerfile. The `php5-pgsql` connection is now working for me.

volker-raschek:
Okay,
I would like to test the container in my environment. I get a feedback tomorrow.

It would be nice, if the user @postfixadmin has an own repository with all docker source code.

Volker

J0WI:
@DavidGoodwin could you also give a feedback here? I think it's not worth to add SQLite support. Apart fro that the image should be fine.

volker-raschek:
Hey JOWI, 
now it's run in my environment, but I have an improvement. 

It's would be nice, if we can check, if the backend has an existing database model. If not we create it for the postfixadmin version or make an upgrade to this version of postfixadmin.

DavidGoodwin:
Hi - It would be nice if it supported SQLite, as at least that would give us support for all three databases through Docker.

Can you add some sort of basic README.txt to your docker directory - so people don't need to figure out how to switch between the two ? 

e.g. 

# Building 

```sh
docker build --pull --rm -t postfixadmin-image .
```

# Running

```sh
docker run -e POSTFIXADMIN_DB_TYPE=mysqli -e POSTFIXADMIN_DB_HOST=whatever -e POSTFIXADMIN_DB_USER=user -e POSTFIXADMIN_DB_PASSWORD=topsecret -e POSTFIXADMIN_DB_NAME=postfixadmin postfixadmin-image
```

Otherwise, looks good to me. Appears to work (well, I got to the login screen :) )

J0WI:
I just added the dependencies for SQLite. I also added the other PDO extensions for consistency.
But I have no idea how SQLite Databases works. @DavidGoodwin Could you help me with the config and setup?

> It's would be nice, if we can check, if the backend has an existing database model. If not we create it for the postfixadmin version or make an upgrade to this version of postfixadmin.

You can simply visit `/upgrade.php` or `/setup.php` to do this. All "Official Images" I've seen so far have the same behavior.

DavidGoodwin:
Please use 'upgrade.php' to setup the various database(s) - doing this will mean the db changes all stay in one place. You ought to be able to run it (upgrade.php) from the command line.

DavidGoodwin:
(There's nothing to do with sqlite databases - assuming the config values are set -  I think you only need : ```$CONF['database_name'] = '/path/to/postfixadmin.db'``` - then just visit upgrade.php and it should get it's schema populated).

cboltz:
> Please use 'upgrade.php' to setup the various database(s) - doing this will mean the db changes all stay in one place. You ought to be able to run it (upgrade.php) from the command line.

It's probably a better idea to use setup.php - it includes upgrade.php (so it will also to the database setup/upgrade), but it comes with some additional checks. Calling upgrade.php is technically possible, but not officially supported ;-)

And yes, running setup.php from the commandline is supported ;-)

J0WI:
`$CONF['database_name']` is set, but the setup does not work for me with SQLite.

DavidGoodwin:
merging; I'll subsequently add some local changes I have to docker-entrypoint.sh so it falls back to sqlite etc. 

Thank you for taking the time to help add this.

DavidGoodwin:
(press the right button!)

J0WI:
Please also push this to the Docker store / have a look at https://github.com/docker-library/official-images#contributing-to-the-standard-library

skriesch:
I have a question about the security in docker-entrypoint.sh. Is there any reason that you want to use as a default database user for MySQL?

J0WI:
By default, the user specified in the database container is used. `root` is only used as a fallback.
You should create a custom user/database/password as specified in the official MariaDB/MySQL image. Also, you shouldn't expose the database at all or only to localhost.

cboltz:
"should" is a good word to describe the difference between theory and practise ;-)

Having root as default makes it too easy to use that - and I probably don't need to mention that you should avoid using root if a non-privileged user is enough. So please don't provide insecure defaults, even if "you should create a custom user". 

People are lazy, so the default should be secure - maybe a "postfixadmin" user?

J0WI:
I did this to be consistent to other containers from the standard library (see e.g. Drupal, Joomla or Wordpress). Linking containers is a deprecated feature, so in future you may be forced to define your environment variables anyway.

DavidGoodwin:
Don't confuse ...

 * running a MySQL/PostgreSQL database container for a single purpose (postfixadmin) and using the "root" credentials for it, with 
 * using root credentials from a shared database server (with many databases etc)

Within a containerised environment, it should be very difficult (and perhaps eventually impossible) to escape the containers 'chroot'. 

See e.g. https://github.com/docker-library/wordpress/blob/f7acf35a2b49b7a45d71b4c5fc8797042f34d4bd/php5.6/apache/docker-entrypoint.sh#L85

skriesch:
I have contributed to a (secure) Wordpress docker image. Look how we are doing it: https://github.com/owncloud-docker/wordpress/blob/master/rootfs/usr/local/bin/wordpress

Our docker image is made for shared services (separate database server). The solution is docker-compose with the user name. You can't use root for localhost on a MySQL server any more. That's fixed since January.

DavidGoodwin:
@skriesch thanks - I hadn't tried the latest MySQL recently ... it does seem you can use root locally - but you have to specify a password.

Would ... 

 a. Defaulting DB_TYPE to sqlite, and
 b. If DB_TYPE != sqlite, require USER/PASSWORD/HOST as env vars ?


https://github.com/docker-library/mysql/blob/86431f073b3d2f963d21e33cb8943f0bdcdf143d/8.0/docker-entrypoint.sh#L82

skriesch:
I recommend using the DB_TYPE as env variable, too. Your proposal is the step into the right direction.
Write down which env vars will be used for which database type. After that you can create something like we have created for ownCloud with env vars: 
https://github.com/owncloud-docker/server/blob/master/docker-compose.yml 

The difference is your support for different database types and docker-compose is for shared systems (like you told). You can use if-else for different env variables and database types then. Do it like you described.

DavidGoodwin:
I pushed https://github.com/postfixadmin/postfixadmin/commit/ba47f2df2ab18453fa46271372033dcd9a66440a earlier .... that removes the linking support and requires the variables be set (else it falls back to a throwaway sqlite db).

skriesch:
That's looking like it should be for multiple database types.

J0WI:
@DavidGoodwin could you please publish the Image on Docker hub?
See also https://github.com/postfixadmin/postfixadmin/pull/64#issuecomment-317248172

J0WI:
@DavidGoodwin could you move the Docker stuff into a dedicated repo? My goal is still to bring this to the official Docker library.
I can provide a FPM and Alpine Linux variant but the folder structure should be adjusted first.

DavidGoodwin:
I can create a new repo under postfixadmin/ for it, sure ... I'm happy to give you write access for that.

I don't quite understand why you'd want to move the docker stuff into it's own repo - do other projects have dedicated docker repos?

J0WI:
You may have a look at the repos linked on https://github.com/docker-library/official-images/tree/master/library
AFAIK all "official" images live on their own repo.

J0WI:
Thanks! I had to make an evil force push to save the history of the past commits in https://github.com/postfixadmin/docker
I've re-added the license in https://github.com/postfixadmin/docker/commit/491dbd6c518d0bd2c0f68464327f959a152a40e9

The removal of Docker from the main repo is here: https://github.com/postfixadmin/postfixadmin/pull/121

PR #70, opened 2017-08-31
virtualmaildel.php with PHP mysqli extension.

Seitanas:
Hello,
this patch enables `virtualmaildel.php` to work with new PHP `mysqli` extension. The old `mysql` extension was deprecated in PHP 5.5.0, and it was removed in PHP 7.0.0.
Have a nice day.


Seitanas:
Hello, could you please change a typo in my last name. Managed to mistype somehow.
Mistyped name is `Ustianvičius` should be `Ustinavičius`.
Thank you.

DavidGoodwin:
done

PR #79, opened 2017-09-26
Harden password reset process

PowerKiKi:
This a follow up to https://github.com/postfixadmin/postfixadmin/pull/18#issuecomment-324013016. I restored old commits, rebased and squashed them all into a single one.

The improvements are:

- Die with an explicit message when a user is trying to reset his lost password and the option is disabled in config
- Redirect user to main page after password change using relative URL
- Don't leak info whether user exists or has recovery info defined
- Throttle password reset requests to prevent brute force attacks
- Show phone/alt email fields in mailbox/admin edit form only when the password reset option is enabled
- Make database upgrade code compatible with other databases types
- Use the existing password generator to generate OTP. It is now stored in database, unique to each user, valid only for 1 hour and can only by used once.


PowerKiKi:
@cboltz, @DavidGoodwin I just rebased this PR. Could you please have a look before it get stale again ?

cboltz:
Thanks for taking care of this, and sorry that I didn't have time to review this earlier - the last two weeks were too busy :-(

Your merge request fixes lots of things (including the security issues the old version of "lost password" had), so I'll merge it now. 

There are some small issues left. For example, things break at various places if a database prefix is used. If you use one of the high level `db_*()` functions like `db_update()`, they already call `table_by_key()` internally. If you hand over a `table_by_key()`-"escaped" table name to them, you'll end up with `$prefix_$prefix_$table` ;-) (my test setup has a `$CONF['database_prefix'] set, and it's not the first time I see such breakage ;-)
I'll see if I can find some time to fix those issues. If not, I'll add some comments here later ;-)

Issue #80, opened 2017-09-26
Enforce code style ?


PowerKiKi:
Would you accept a PR that use [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and Travis CI to enforce a consistent code style throughout the entire PHP code base ?

I believe this would make it easier to contribute, by avoiding a lots of annoying useless diff (such as trailing whitespaces) and by automating some feedback on incoming PR.

If you agree on the principle, I could suggest PHP-CS-Fixer rules, or you choose them yourself and let me implement the rest via composer and Travis.

DavidGoodwin:
Sounds good to me. Please submit a pull request when you are able to.

PR #83, opened 2017-10-09
Introduce PHP-CS-Fixer and Travis to enforce code style

PowerKiKi:
Commands `composer check` and `composer fix` can be used to respectively
check and automatically fix code style in the entire project.

This is a follow-up to #80. There is a huge amount of code changes, but it's strictly code style automatically applied by PHP-CS-Fixer, no logic was changed. You can have a look at the rules in `.php_cs.dist` and tweak them if needed and then apply them with `composer fix`. We stayed very conservative with the rule choices, specifically avoid all _risky_ rules, to be confident that no bugs were introduced.

When merging this PR, you will have to enable Travis for this repository. Let me know if you need help for that.

Also I declared PHP 5.2 as requirement for this project in `composer.json`, based on the instructions found in `INSTALL.TXT`. But IMHO this should be upgraded to at least PHP 5.6 because [it is the last supported PHP version in the 5.X branch](http://php.net/supported-versions.php). But that's out of topic.


DavidGoodwin:
looking at the config file you attached, it looks like there are loads of options to look at.

Why not just ```--rules=@PSR2``` ? 

If we're going to change the indentation/formatting of most of the code, I think we should just adopt a common standard, rather than making our own up....

DavidGoodwin:
a. Upgrading to PHP 5.6+ is overdue.
b. I think I'd rather the ruleset was more simple - i.e. following PSR2 or whatever. If we're going to reformat most of the code we might as well follow something that's a standard rather than making our own.

therefore I think I'd prefer a .php_cs.dist  file like : 

```<?php

$finder = PhpCsFixer\Finder::create()
    ->exclude('smarty')
    ->exclude('vendor')
    ->exclude('model') 
    ->exclude('templates')
    ->exclude('templates_c')
    ->exclude('debian')
    ->in(__DIR__);

return PhpCsFixer\Config::create()
    ->setFinder($finder)
    ->setRules(array('@PSR2'=> true));
```

DavidGoodwin:
(I excluded 'model' as phpcs sometimes seems to create invalid syntax in the files; I'm not sure anyone except cboltz understands what's going on in there anyway).

PowerKiKi:
The suggested configuration (with the new commit) is entirely PSR2 compliant.

But because PSR2 does not cover everything, selecting only the PSR2 rules may not be enough. So we add more things in addition of PSR2 to be more _strict_. Many of those things comes from Symfony rules.

If you want to convince yourself, you can try to re-run the tool with ony PSR2. You will see that nothing changes.

Could you elaborate on the syntax issue in `model/` ? The whole point of this is to standardize the code. If we only apply the rules to half of it, it loses a lot of interest.

DavidGoodwin:
Why is there a long list of options in .php_cs.dist? 

What's the advantage of specifying all those rules over just having a @PSR2 => true?

DavidGoodwin:
model/AliasHandler.php gets a syntax error in the mailbox_goto column. Running ```for x in model/* ; do php -l $x ; done ``` highlights it.

Additionally, I suspect @cboltz will cry when he loses the spacing within his various pacol() definitions.

I see no need to be 'more strict' - I'd rather we just followed a universal standard than creating our own. If we're PSR2 compliant, people submitting patches are more likely to be correct. For every extra option you add to phpcs, it's one more hurdle to contributions... surely?

PowerKiKi:
> model/AliasHandler.php gets a syntax error

I don't get anything with PHP 5.6 and 7.1. What PHP version are you using ? And what line is invalid ?

```sh
$ find model/ -type f -exec php5.6 -l {} \;
No syntax errors detected in model/CliEdit.php
No syntax errors detected in model/CliScheme.php
No syntax errors detected in model/CliHelp.php
No syntax errors detected in model/PFAHandler.php
No syntax errors detected in model/DomainHandler.php
No syntax errors detected in model/AdminHandler.php
No syntax errors detected in model/CliView.php
No syntax errors detected in model/AliasdomainHandler.php
No syntax errors detected in model/VacationHandler.php
No syntax errors detected in model/FetchmailHandler.php
No syntax errors detected in model/CliDelete.php
No syntax errors detected in model/Config.php
No syntax errors detected in model/AliasHandler.php
No syntax errors detected in model/MailboxHandler.php
No syntax errors detected in model/AdminpasswordHandler.php
```

> @cboltz will cry when he loses the spacing within his various `pacol()`

[PSR2 forbids this style of formatting](http://www.php-fig.org/psr/psr-2/#44-method-arguments). It must have exactly 1 space after a comma for argument.

> for every extra option [...] it's one more hurdle to contributions

On the contrary, it's one thing less to wonder about and argue over. These are a very clear rules that can be applied automatically (virtually no extra work), that eliminate doubts, inconsistencies. I'd even argue that it boost productivity because the code style is very consistent within the project itself, but across the entire PHP ecosystem. Those rules are commonly used by PHP-CS-Fixer users and should be well-known in the community. And they will most likely align with the incoming PSR-12.

On my projects, I would never consider less automated formatting rules is better. But here you get to decide, I'll follow whatever you decide.

DavidGoodwin:
```PHP CS Fixer 2.6.0 Local Tavern by Fabien Potencier and Dariusz Ruminski (5642a36)```

and PHP from deb.sury.org -

```PHP 7.0.24-1+0~20170928163718.7+stretch~1.gbp4e7e8f (cli) (built: Sep 28 2017 16:37:19) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.24-1+0~20170928163718.7+stretch~1.gbp4e7e8f, Copyright (c) 1999-2017, by Zend Technologies
    with Xdebug v2.5.1, Copyright (c) 2002-2017, by Derick Rethans
```
....


```
$ php -l model/AliasHandler.php 
PHP Parse error:  syntax error, unexpected ',' in model/AliasHandler.php on line 85
Errors parsing model/AliasHandler.php
```

```php
            'goto_mailbox'  => pacol(
                $mbgoto,
                $mbgoto,
                ,
                ,
                'pEdit_alias_forward_and_store' ,
                ''                                ,
                0,
                /*options*/ '',
                /*not_in_db*/ 1
            ), # read_from_db_postprocess() sets the value
```

which was originally : 

```
            'goto_mailbox'  => pacol(   $mbgoto,    $mbgoto,$mbgoto,'bool', 'pEdit_alias_forward_and_store' , ''                                , 0,
                /*options*/ '',
                /*not_in_db*/ 1                         ), # read_from_db_postprocess() sets the value
```

PowerKiKi:
On my latest commit, the [line 85 do have a `$mbgoto`](https://github.com/Ecodev/postfixadmin/blob/0a0278d0d68345932dc53fa644fb3fafde2754bf/model/AliasHandler.php#L85). So there is no syntax issue.

Also [Travis successfully check those rules on PHP 5.6, 7.0 and 7.1](https://travis-ci.org/Ecodev/postfixadmin). And locally I can successfully check and fix without producing any changes:

```sh
$ rm -f .php_cs.cache && composer check && composer fix && git status
> php-cs-fixer fix --ansi --dry-run --diff
Loaded config default from "/sites/3dparty/postfixadmin/.php_cs.dist".

Checked all files in 7.162 seconds, 14.000 MB memory used
> php-cs-fixer fix --ansi
Loaded config default from "/sites/3dparty/postfixadmin/.php_cs.dist".
Using cache file ".php_cs.cache".

Fixed all files in 0.008 seconds, 8.000 MB memory used
On branch code-style
Your branch is up-to-date with 'origin/code-style'.
nothing to commit, working directory clean
```

This is probably because you seem to be using a global PHP-CS-Fixer 2.6, instead of the locally installed 2.7.1 one. Try `composer fix` or `./vendor/bin/php-cs-fixer` instead of whatever you are using.

DavidGoodwin:
I just used what the phpcs website said to download ....  it seems to behave using 2.7.1...

PowerKiKi:
You should use composer to install correct dependencies instead of using global install.

Also let me know what you decide about the rules. Only PSR2, or more strict ?

PS: `phpcs` usually refer to [PHP_CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer), not PHP-CS-Fixer

cboltz:
I had a quick look at the changes the code formatting would do. Some comments:

* is it possible to convert `//` comments to `#` instead of doing it the other way round? We have `//` for historic reasons, but I never liked it, and new code has `#` ;-)

* it seems the formatting removes conditions that can never be false, see for example ADDITIONS/squirrelmail-plugin/functions.inc.php around line 50 where the `if` branch calls `exit()` and the following `else` gets dropped (and the code that was inside the `else` gets executed unconditionally). That's not really wrong, but more than I'd expect from "formatting" ;-) However, I sometimes prefer to keep those conditions as safety net in case the `exit()` gets dropped from the `if` branch.

* I'd don't like to have the opening brace in functions and classes in a separate line. I'd prefer to have it as one `function foo() {` line, similar to what you do for `if` blocks.

* composer.lock says the file is autogenerated. Does it make sense to add it to git, or should we add it to .gitignore instead?

* as David already wrote, please do not change whitespace between array items or function parameters. We have several places with "table-style" code, and while this might not match your usual coding style, it makes the code much more readable. Just some examples:
    * config.inc.php (for example around line 190)
    * edit.php (around line 110)
    * model/DomainHandler `delete()` (around line 221 / 347) 
    * upgrade.php (the $replace arrays)
    * Oh, and `initStruct()` in model/*Handler is completely unreadable with the formatting you propose. But I have to correct David's expectations here - I won't cry, I'll simply revert anything that breaks this whitespace \*eg\*

* another whitespace thing - if there are two empty lines, there might be a reason for it (like grouping some functions). I'm not sure if reducing everything to only one empty line helps.

* in model/PFAHandler, several variables are initialized as `= null`. This `= null` gets dropped. I know `null` is the default, but I'd still prefer to have that in the code so that it's obvious.

* you might also have a look at https://sourceforge.net/p/postfixadmin/wiki/Coding_guidelines/ which includes a few coding styles - and yes, I'm aware that we still have quite some old code that doesn't even follow them for historical reasons

* final note: once we accept this pull request, we'll probably break any other open pull request. Re-creating automated changes is much easier ;-) therefore we should first handle (most of) the open pull requests before merging this one.

PowerKiKi:
> convert `//` comments to `#`

The rule `hash_to_slash_comment` can only enforce `//`, not anything else. So I disabled it altogether, but it goes against an extremely popular style in PHP ecosystem.

> `else` gets dropped 

I disabled the rule `no_useless_else`.

> opening brace in functions

I re-configured the rule `braces`.


> composer.lock in .gitignore ?

`composer.lock` must be commited, as [explained in the official documentation](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control).

> do not change whitespace between array items or function parameters

I disabled the rule `no_whitespace_before_comma_in_array`, `trim_array_spaces` and `method_argument_space`, and adjusted `binary_operator_spaces`, but this [goes against PSR2](http://www.php-fig.org/psr/psr-2/#44-method-arguments).


> two empty lines

I disabled `no_extra_consecutive_blank_lines`.

> The `= null` gets dropped

I disabled `no_null_property_initialization`.

> https://sourceforge.net/p/postfixadmin/wiki/Coding_guidelines/ 

I was not aware of that, thank you. It seems thoses guidelines are now respected.

> should first handle (most of) the open pull requests

Absolutely yes. Luckily there are not so many nowadays.


I rebased the PR, and split it into two commits. The first one defines the rules, and the second commit apply them (automatically). That way it will be easier to cherry-pick the rule, and re-apply all of them on the current code base whenever the PR is merged.

DavidGoodwin:
I'd still rather we just adopted a minimal PSR2 - then most editors will just format it correctly and contributors won't need to do anything. 

 * I don't see any point in switching to/from # and // ... just leave them as is.
 * config.inc.php should be left unaltered if at all possible ... any change to it will annoy anyone who has installed the .deb

PowerKiKi:
Force pushed the exclusion of `config.inc.php`.

At this point, if you want further modifications, I'll let you debate between you two first. Because @DavidGoodwin says PSR2 only, but @cboltz says things against PSR2 and seem to agree with other non-PSR2 changes. Personally I'd merge it as is, because it now is a compromise between what you both want.

Once you reach a conclusion, let me know. Or you could also go ahead and edit it further yourself, I think you have a pretty good basis now.

DavidGoodwin:
A minimal php_cs.dist and the travis.yml things have been merged. Other changes to master will have messed up merging of this. It's now just a case of someone having hte courage to run the reformat and commit it.

Thanks for taking the time to help with the above @PowerKiKi

PR #84, opened 2017-10-11
Php cs reformatting

DavidGoodwin:
this would replace https://github.com/postfixadmin/postfixadmin/pull/83 

main change : simpler php-cs config - PSR2 with one difference.


DavidGoodwin:
partly merged (without the reformatting); closing for now.

PR #85, opened 2017-10-12
fix invalid value for token_validity

julywind:
failed to create a mailbox， because empty string  is not a valid datetime value.

### DIFF
--- a/model/AdminHandler.php
+++ b/model/AdminHandler.php
@@ -79,7 +79,7 @@ class AdminHandler extends PFAHandler {
             'phone'           => pacol( 1,  $passwordReset, 0,      'text', 'pCreate_mailbox_phone', 'pCreate_mailbox_phone_desc', ''),
             'email_other'     => pacol( 1,  $passwordReset, 0,      'mail', 'pCreate_mailbox_email', 'pCreate_mailbox_email_desc', ''),
             'token'           => pacol( 1,          0,      0,      'text', ''                   , ''                  ),
-            'token_validity'  => pacol( 1,          0,      0,      'ts',   ''                   , ''                  ),
+            'token_validity'  => pacol( 1,          0,      0,      'ts',   ''                   , '', '2000-01-01 00:00:00'),
             'created'         => pacol( 0,          0,      0,      'ts',   'created'            , ''                  ),
             'modified'        => pacol( 0,          0,      1,      'ts',   'last_modified'      , ''                  ),
         );
--- a/model/AdminHandler.php
+++ b/model/AdminHandler.php
@@ -79,7 +79,7 @@ class AdminHandler extends PFAHandler {
             'phone'           => pacol( 1,  $passwordReset, 0,      'text', 'pCreate_mailbox_phone', 'pCreate_mailbox_phone_desc', ''),
             'email_other'     => pacol( 1,  $passwordReset, 0,      'mail', 'pCreate_mailbox_email', 'pCreate_mailbox_email_desc', ''),
             'token'           => pacol( 1,          0,      0,      'text', ''                   , ''                  ),
-            'token_validity'  => pacol( 1,          0,      0,      'ts',   ''                   , '', '2000-01-01 00:00:00'),
+            'token_validity'  => pacol( 1,          0,      0,      'ts',   ''                   , '', date("Y-m-d H:i:s",time())),
             'created'         => pacol( 0,          0,      0,      'ts',   'created'            , ''                  ),
             'modified'        => pacol( 0,          0,      1,      'ts',   'last_modified'      , ''                  ),
         );
--- a/model/MailboxHandler.php
+++ b/model/MailboxHandler.php
@@ -42,7 +42,7 @@ class MailboxHandler extends PFAHandler {
             'phone'         => pacol(   1,  $passwordReset, 0,      'text', 'pCreate_mailbox_phone'         , 'pCreate_mailbox_phone_desc'       , ''),
             'email_other'   => pacol(   1,  $passwordReset, 0,      'mail', 'pCreate_mailbox_email'         , 'pCreate_mailbox_email_desc'       , ''),
             'token'         => pacol(   1,          0,      0,      'text', ''                              , ''                                 ),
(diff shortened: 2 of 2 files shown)


DavidGoodwin:
show create table mailbox : 
```sql
....
  `token` varchar(255) CHARACTER SET utf8 NOT NULL DEFAULT '',
  `token_validity` datetime NOT NULL DEFAULT '2000-01-01 00:00:00',
```

Is your database schema correct? Are you using PostgreSQL by any chance?

DavidGoodwin:
OK, so I get a value for mailbox.token_validity of 0000-00-00 00:00:00 .. which probably isn't what is expected. I don't like hard coding 2000-01-01 in the code. 

Ideally it should be set to something like NOW() ...

db_insert() in functions.inc.php has a list of columns which can be timestampable - 

```
function db_insert ($table, $values, $timestamp = array('created', 'modified') ) {
```
and defaults to created/modified. 

Ideally, MailboxHandler etc should correctly call ```db_insert()``` correctly. 

I suspect to fix that, it's a case of all the Handler classes being aware of what timestampable fields they're responsible for (some sort of protected property) that value then being used in PFAHandler (the parent class).

@cboltz do you need to insert 2p here?

cboltz:
Actually we already hardcode 2000-01-01 in the database scheme because MySQL doesn't allow CURRENT_TIMESTAMP in more than one column. For the 'created' and 'modified' fields, we have NOW() in place by default (IIRC that's the default set db_insert() uses).

If you really want to initialize this field as NOW(), that would need a few changes in PFAHandler. It would need to convert an empty default for a 'ts' field to NOW(), and it would also make sure to do that for 'created' and 'updated'. That's in theory possible, but in practise just going with 2000-01-01 is the easier solution.

I won't object if someone implements this in PFAHandler, but for now, using the 2000-01-01 default is the easiest solution. Actually a default expire date in the past for the recovery token might be seen as a good thing ;-)

model/AdminHandler.php
```diff
@@ -79,7 +79,7 @@ protected function initStruct() {
             'phone'           => pacol( 1,  $passwordReset, 0,      'text', 'pCreate_mailbox_phone', 'pCreate_mailbox_phone_desc', ''),
             'email_other'     => pacol( 1,  $passwordReset, 0,      'mail', 'pCreate_mailbox_email', 'pCreate_mailbox_email_desc', ''),
             'token'           => pacol( 1,          0,      0,      'text', ''                   , ''                  ),
-            'token_validity'  => pacol( 1,          0,      0,      'ts',   ''                   , ''                  ),
+            'token_validity'  => pacol( 1,          0,      0,      'ts',   ''                   , '', '2000-01-01 00:00:00'),
```
cboltz: Please add some whitespace (spaces, no tabs) so that the new default actually starts in the "column" meant for the default.

model/MailboxHandler.php
```diff
@@ -42,7 +42,7 @@ protected function initStruct() {
             'phone'         => pacol(   1,  $passwordReset, 0,      'text', 'pCreate_mailbox_phone'         , 'pCreate_mailbox_phone_desc'       , ''),
             'email_other'   => pacol(   1,  $passwordReset, 0,      'mail', 'pCreate_mailbox_email'         , 'pCreate_mailbox_email_desc'       , ''),
             'token'         => pacol(   1,          0,      0,      'text', ''                              , ''                                 ),
-            'token_validity'=> pacol(   1,          0,      0,      'ts',   ''                              , ''                                 ),
+            'token_validity'=> pacol(   1,          0,      0,      'ts',   ''                              , '',	'2000-01-01 00:00:00'                        ),
```
cboltz: There's a tab before the default value. Please convert it to spaces and ensure the default value is in the correct "column".

Also, please remove the whitespace in front of the closing parenthesis.

quicktrick:
@cboltz   After I made the changes in AdminHandler.php and MailboxHandler.php you mentioned in above messages, I got the following errors when trying to add a new mailbox:

> 2017/11/20 15:04:04 [error] 23954#23954: *3547 FastCGI sent in stderr: "PHP message: Config::read(): attempt to read undefined config option "__LANG.2000-01-01 00:00:00", returning null" while reading response header from upstream, client: 192.168.1.1, server: mydomain.tld, request: "GET /postfixadmin/edit.php?table=mailbox&domain=mydomain.tld HTTP/2.0", upstream: "fastcgi://unix:/run/php/php7.1-fpm.sock:", host: "mydomain.tld", referrer: "https://mydomain.tld/postfixadmin/list-virtual.php"

> 2017/11/20 15:04:11 [error] 23954#23954: *3547 FastCGI sent in stderr: "PHP message: Config::read(): attempt to read undefined config option "__LANG.2000-01-01 00:00:00", returning null" while reading response header from upstream, client: 192.168.1.1, server: mydomain.tld, request: "GET /postfixadmin/edit.php?table=mailbox&domain=mydomain.tld HTTP/2.0", upstream: "fastcgi://unix:/run/php/php7.1-fpm.sock:", host: "mydomain.tld", referrer: "https://mydomain.tld/postfixadmin/list-virtual.php"

> 2017/11/20 15:04:31 [error] 23954#23954: *3547 FastCGI sent in stderr: "PHP message: Config::read(): attempt to read undefined config option "__LANG.2000-01-01 00:00:00", returning null PHP message: Invalid query: Incorrect datetime value: '' for column 'token_validity' at row 1 PHP message: caused by query: INSERT INTO mailbox (username,local_part,domain,maildir,password,name,quota,active,phone,email_other,token,token_validity,created,modified) VALUES ('test@mydomain.tld','test','mydomain.tld','mydomain.tld/test/','$1$bf3922bc$TQjFfgOJciICNVLVuSZpg.','test','2097152000','1','','','','',now(),now())" while reading response header from upstream, client: 192.168.1.1, server: mydomain.tld, request: "POST /postfixadmin/edit.php?table=mailbox&domain=mydomain.tld HTTP/2.0", upstream: "fastcgi://unix:/run/php/php7.1-fpm.sock:", host: "mydomain.tld", referrer: "https://mydomain.tld/postfixadmin/edit.php?table=mailbox&domain=mydomain.tld"

PR #90, opened 2017-11-11
singular correction.  Update en.lang

pbkwee:
"Your email address or password are not correct".  Message indicates one thing is wrong.  So use is not are.


DavidGoodwin:
Thanks; merged.

Issue #91, opened 2017-11-14
Support MySQL over SSL


leeclemens:
postfixadmin should allow the ability to configure backend MySQL connections over an SSL connection, using the standard functionality provided by mysql client and php

DavidGoodwin:
I think this would be best supported by supporting PDO.....

leeclemens:
The PR is based on the pre-existing use of mysqli and merely extends it to accept SSL related flags.
Are you suggesting all of the MySQL functionality be replaced in favor of PDO? (I do not claim to be a PHP expert, so forgive me if I am misunderstanding something.)

DavidGoodwin:
Oh - sorry - I was being blind - I didn't spot the PR for some reason.

I've never used MySQL with SSL, so I can't say much about your code ..... it /looks/ alright !

leeclemens:
Ahh, no worries!  I emailed a patch about a year ago similar to this PR and had been using it prior to writing this PR. Using this specific code now without issue for a few days.

I suppose if this works as-is with non-SSL MySQL connections and fails when `database_use_ssl` is set to `true` without SSL configured with MySQL it could be considered fairly reliable?

It really just passes arguments to `mysqli_real_connect`, and I admit I have not tested all combinations - but those are handled by PHP and MySQL once passed in.

DavidGoodwin:
Hi - I'll merge this - but really postfixadmin should move to use PDO - and things like SSL flags etc would then be handled through whatever the PDO connection string / options are.

PR #92, opened 2017-11-15
Add support for MySQL connections over SSL

leeclemens:
Allows for the configuration of MySQL SSL related options and passes them accordingly to the underlying library.

Fixes: #91


DavidGoodwin:
Thank you for contributing!

PR #94, opened 2017-11-20
Add bcrypt password option

adjoeu:
This would allow to use bcrypt through php's [password_hash() ](http://php.net/manual/en/function.password-hash.php) function while keeping dovecot compatibility

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -161,6 +161,8 @@ $CONF['smtp_client'] = '';
 //     - don't use dovecot:* methods that include the username in the hash - you won't be able to login to PostfixAdmin in this case
 //     - you'll need at least dovecot 2.1 for salted passwords ('doveadm pw' 2.0.x doesn't support the '-t' option)
 //     - dovecot 2.0.0 - 2.0.7 is not supported
+// bcrypt = blowfish encrypted through php's password_hash(). Example: bcrypt:14:{BLF-CRYPT}:2y:2a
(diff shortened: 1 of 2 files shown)


bakelnaar:
I've checked the code out. It seems to be working okay, however I needed to change the following line in my configuration file:
$CONF['encrypt'] = 'bcrypt:14::2y:2a'; instead of $CONF['encrypt'] = bcrypt:14:{BLF-CRYPT}:2y:2a
The second option (which is in the documentation) adds {BLF-CRYPT} to each hash, which makes it impossible to login again in PFA with that user account (or change the password).
I implemented this pull request in the PFA 3.1 release, so it also could be that I miss some other change in the code, although I couldn't find any with a quick look at the network graph.

adjoeu:
Could you verify that the preg_match works as intended? I get:
`$matches = Array ( [0] => {BLF-CRYPT}$2a$14$y4gIOX... [1] => {BLF-CRYPT} [2] => $2a) `
and then verifies against
`$vrf_storedPassword = $2y$14$y4gIOX... `

As a test i removed {BLF-CRYPT} directly within the database (without changing the configuration) 
`Array ( [0] => $2a$14$y4gIOX... [1] => [2] => $2a) `
and again verifies against
`$2y$14$y4gIOX...`

adjoeu:
this may be closed. the requested feature was added with 3.2

Issue #109, opened 2017-12-19
syslog ident


Erwane:
Hi,

could you set the "syslog ident" to vacation instead of script path ?

DavidGoodwin:
merged....

PR #110, opened 2017-12-19
#109 : syslog ident set to "vacation"

Erwane:


DavidGoodwin:
thanks for contributing!

PR #112, opened 2017-12-24
Update functions.inc.php

er1cs:
I found that Mysql 8 don't like table names without `` in requests. So i make changes in function table_by_key in functions.inc.php and in upgrade.php . Now it works.  FreeBSD 11.1 Apache/2.4.29 (FreeBSD) PHP/7.1.11 Mysql 8

### DIFF
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -1802,7 +1802,7 @@ function table_by_key ($table_key) {
         $table = $CONF['database_tables'][$table_key];
     }
 
-    return $CONF['database_prefix'].$table;
+    return $CONF['database_prefix']."`".$table."`";
 }
 
 /*


DavidGoodwin:
Thank you!

niallor:
Unfortunately, a back-tick-quoted table name seems to be invalid syntax for PostgreSQL.

[Wed Jan 03 14:14:32.677862 2018] [:error] [pid 31784] [client 2a04:2e00:1:1::a:42758] Invalid query: ERROR:  syntax error at or near "`"\nLINE 1: CREATE TABLE  `config` ( \n                      ^
[Wed Jan 03 14:14:32.677934 2018] [:error] [pid 31784] [client 2a04:2e00:1:1::a:42758] caused by query: CREATE TABLE  `config` ( \n                    id SERIAL,\n                    name VARCHAR(20) NOT NULL UNIQUE,\n                    value VARCHAR(20) NOT NULL,\n                    PRIMARY KEY(id)\n                    )

PR #113, opened 2017-12-24
Update upgrade.php

er1cs:
I found that Mysql 8 don't like table names without `` in requests. So i make changes in function table_by_key in functions.inc.php and in upgrade.php . Now it works.  FreeBSD 11.1 Apache/2.4.29 (FreeBSD) PHP/7.1.11 Mysql 8


DavidGoodwin:
Thank you!

PR #116, opened 2018-01-09
Ja fix

mzch:
Update ja.lang


DavidGoodwin:
Thanks for taking the time to contribute ! :) :+1:

cboltz:
Thanks for the translation update!

May I ask you for a small follow-up commit?

The `# XXX` comments are meant as markers on texts that need translation or an update - so after translating a text, please remove these comments. This is especially important for changed texts because only people who understand a language can judge if the translation needs to be updated (I know the difference between english and japanese, but don't understand japanese ;-)

Oh, and `language-update.sh --stats` uses the `# XXX` comments, so if you don't remove them, it will think there are lots of untranslated strings ;-)

mzch:
done! ;)

DavidGoodwin:
(merged)

PR #122, opened 2018-01-19
Typo correction

dryware-fr:
There was a typo in the script, rendering a TCP connection to MySQL or Postgres impossible.


DavidGoodwin:
Thanks for contributing @dryware-fr

PR #124, opened 2018-01-22
Password reset  - fix invalid DB query (double backtick)

sitilge:
When restoring admin's password, double backtick results in a query error

> PHP message: Invalid query: Incorrect table name ''
> PHP message: caused by query: SELECT * FROM ``admin`` WHERE username='sitilge@example.com'"


DavidGoodwin:
Thanks for contributing!

PR #131, opened 2018-02-06
Destroy session when loading user login page

### LINKED ISSUE #130: Logout does not work
Steps to reproduce:
- Login (either as admin or user)
- Click on the "logout" link which opens the corresponding "login.php" page
- User is still logged in (shown in footer and able to simply open "main.php" again)

Expected behavior:
- User is logged out, i.e., session is invalidated

Version affected:
- Master (tested with 220f828)

Please let me know if I can provide any further info.

cygery:
Fixes #130

### DIFF
--- a/public/users/login.php
+++ b/public/users/login.php
@@ -58,6 +58,11 @@ if ($_SERVER['REQUEST_METHOD'] == "POST") {
     }
 }
 
+
+session_unset();
+session_destroy();
+session_start();
+
 $_SESSION['PFA_token'] = md5(uniqid(rand(), true));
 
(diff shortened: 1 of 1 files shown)


DavidGoodwin:
Thanks!

gnanet:
Hi, only a short comment, because i faced this recently, and altough the session was not known by the server, the sessionID in the browsers cookie was presen in the sessions-folder of PHP and the user was logged back every time you logged out.

It is still good to check if any session is known to the server with  `if (isset($_SESSION))`, and "start" a session before calling session_unset + session_destroy

PR #142, opened 2018-02-22
Fixed "Incorrect integer value: 'Array' for column" error in updates.

racerxdl:
Fixed a small typo that causes update not to work.


DavidGoodwin:
Thanks for contributing :)

PR #145, opened 2018-02-28
Add predefined quota on mailbox creation page

o-m-d:
Filling Quota on mailbox creation page with value, defined in $CONF['mailbox_default_quota'].
It's useful for creating mailboxes with same quota.
When  $CONF['mailbox_default_quota'] is empty (by default), mailbox creation page works as before - with empty Quota.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -321,6 +321,8 @@ $CONF['quota'] = 'NO';
 $CONF['domain_quota'] = 'YES';
 // You can either use '1024000' or '1048576'
 $CONF['quota_multiplier'] = '1024000';
+// Quota for create mailbox page
+$CONF['mailbox_default_quota'] = '';
 
 // Transport
 // If you want to define additional transport options for a domain set this to 'YES'.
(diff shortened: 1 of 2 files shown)


cboltz:
I agree that the quota field should be pre-filled (actually I wonder why it isn't and would call that a regression).

However, I don't agree with introducing a new config option for this - we already have enough config options ;-) and this isn't a case of "one size fits all [domains]". Maybe using the maximum allowed quota as default.would make sense? Typical users don't want less anyway ;-)

Hint: updateMaxquota() updates the description text with the maximum allowed quota, also updating the default there should be easy.

o-m-d:
In some cases (eg on VPS or outdated hardware (sad but true)) mail admins don't want grant max quota to all mailboxes, and also want pre-filled some quota at every mailbox creation. In that cases I'm apply this patch manually.

I'm agree - one size to all domains not best solution. Maybe better add this as per-domain option, like a maxquota option? But it's requires more time.

lelutin:
per-mailbox max quota would be a good value to pre-fill the field with imho. if I understand correctly there is already this value available, either as the default per-mailbox setting in $CONF['maxquota'], or set per-domain (probably in the database)

PR #148, opened 2018-03-08
Don't reply to known autoresponses

csware:
CommunigatePro and Outlook use different headers to indicate AutoResponses, handle those correctly.


DavidGoodwin:
Thanks for contributing!

PR #157, opened 2018-03-27
Added configuration hook for PFA.login function and sample

motwok:
The basic idea was to allow the usage of 2FA and OTP PAM providers with postfixadmin.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -180,6 +180,27 @@ if(file_exists('/usr/bin/doveadm')) {
     $CONF['dovecotpw'] = "/usr/bin/doveadm pw"; # debian
 }
 
+// If you want to define your own function to verify the login set this to the name of the function.
+// Notes: 
+//   - this configuration directive will override PFAHandler.login
+//   - the login_hook() function example is present below, commented out
+//   - if the function does not exist the program will default to the normal login function
+$CONF['login_hook'] = 'NO'
+
+/*
+    login_hook example function
+ 
+    Called when login is needed if $CONF['login_hook'] == '<name_of_the_function>'
+    - allows you to use customized login validation like PAM with OTP or 2FA
+    - the example below will use PAM, to use it you must configure PAM (e. g. create /etc/pam.d/php) first
+*/
+/*
+function login_hook($username, $password) {
+    return pam_auth( $username, $password);
+}
+*/
+
+
 // Password validation
 // New/changed passwords will be validated using all regular expressions in the array.
 // If a password doesn't match one of the regular expressions, the corresponding
--- a/model/PFAHandler.php
+++ b/model/PFAHandler.php
@@ -824,6 +824,13 @@ abstract class PFAHandler {
      * @return boolean true on successful login (i.e. password matches etc)
      */
     public function login($username, $password) {
+
+        $login_hook = Config::read('login_hook');
+
+        if ($login_hook != 'NO' && function_exists($login_hook)) {
+            return $login_hook($username, $password);
+        }
+
         $username = escape_string($username);
 
         $table = table_by_key($this->db_table);


cboltz:
As discussed on IRC, I slightly ;-) doubt that we need a hook function to support this.

IMHO extending `pacrypt()` with a `$username` parameter would be a good idea (it's also needed for some dovecot hashing methods).

Afterwards, we could just add a `_pacrypt_pam()` method like the (untested) one below and make sure that `pacrypt()` calls it if `$CONF['encrypt'] = 'pam'` is set.

```
_pacrypt_pam($username, $pw, $pw_db = '') {
    if ($pw_db) {
        # login, validate password
        if p(am_auth($username, $pw) {
            return $pw_db;  # auth successful, caller expects return value == $pw_db
        } else {
            return false;
         }
    } else {
        # hash/set new password
        return md5($pw);  # md5 is obviously a bad idea, but you should get the point ;-)
    }
}
```

DavidGoodwin:
I've thought adding wordpress style filter/hook function calls in the Postfixadmin codebase - which would (I think) support what you're asking for.

Then I think it'd be a case of you doing a ```register_hook('pacrypt', 'my_auth_function');``` or something similar. There wouldn't be any more config options to support this, but people would end up having random bits local to their installation.

motwok:
David, my problem is not the way to hook the function... if you give me an example how to hook PFAHandler.login i will replace it.

But the main-discussion with cboltz was/is about where the functionality has do be located.
cboltz wants this in pacrypt... 
I strongly recommend not to (ab)use pacrypt and put authentication specific stuff into it.
(May pacrypt is used in login but it does there what it should do: crypting the Passwort.)

If we add pacrypt_pam function (configured with the pacrypt parameter from the config), how do we configure what pacrypt_pam is returning if really used for crypting the password? 
Another configuration entry?

DavidGoodwin:
Yes, I know @cboltz wants the functionality merged into the codebase, so anyone can use it. 

And I know he's also against adding more config parameters.


From the conversation I read (quickly) on IRC, I think you're asking that 2FA is just enabled/used on login to PFA itself - and not for end users (to mailboxes). 

PostfixAdmin currently uses the login() function ( https://github.com/postfixadmin/postfixadmin/blob/master/model/PFAHandler.php#L826 ) for authentication to itself (as a web app).

In some imaginary world where I added hooks into the codebase, there would presumably be a pre and post login hook in PFAHandler::login() ... so you'd just register your new auth thing there. Perhaps something horrible like the following ...

``` php
   public function login($username, $password) {
        $username = escape_string($username);

        $table = table_by_key($this->db_table);
        $active = db_get_boolean(true);
        $query = "SELECT password FROM $table WHERE " . $this->id_field . "='$username' AND active='$active'";

        $result = db_query($query);
        if ($result['rows'] == 1) {
            $row = db_assoc($result['result']);
            $crypt_password = pacrypt($password, $row['password']);

            if ($row['password'] == $crypt_password) {
                return true;
            }
        }
       // pacrypt auth failed; perhaps there's a custom login handler registered -
        $retval = apply_hooks('PFAHandler_login', $username, $password);
        if($retval === true) { return true; } 
        return false;
    }```

and then in your code somewhere (perhaps in config.local.php) : 
```php
function my_auth_stuff($username, $password) { 
     return pam_auth($username, $password); 
}
register_hook('PFAHandler_login', 'my_auth_stuff');
```

Or with PHP 5.6+ (?) :
```
register_hook('PFAHandler_login', function($username, $password) { 
    return pam_auth($username, $password); 
}
```


[I'm probably not using / porting how Wordpress's hooks function very well ... I'm not really looked at it's code in detail].

motwok:
David... 

i am sure that this is a nightmare... hooking, injection... you name it.

I can change to integrate pam_auth() in the login... making login itself to _login_sql()...a new _login_pam()... and a new login() function that distributes to the _login_* functions... you get the picture

Unfortunately this needs a config variable...

If agreed i have no problem to change it. 

Emmo

PS: I have 2FA running for Mailbox access... it is a bit anoying for the users but  it works.

motwok:
david, cbolz,

i had a little time where should this "Login_*" functions be implemented... functions.inc.php or inside PFAHandler.php ???

Emmo

DavidGoodwin:
The problems highlighted in this ticket (no hook support) is kind of related to the hacky changes in functions.inc.php's db_insert ( https://github.com/postfixadmin/postfixadmin/blob/d4e5470f7c1551ba20ad601313a2554972a5da27/functions.inc.php#L1884 ) and db_update ( https://github.com/postfixadmin/postfixadmin/blob/d4e5470f7c1551ba20ad601313a2554972a5da27/functions.inc.php#L1952 ) for password_expiration and the 'recaptcha' support in issue #224  ....

PR #167, opened 2018-04-27
vacation: Correctly process mails with recipient_delimiter

csware:
If the recipient address has vacation enabled and a recipient_delimiter (e.g., `+`) is used, then the rewritten address is not correctly restored (e.g., `test#example.com+foo@autoreply.example.com` -> `test#example.com@autoreply.example.com+foo` and not `test@example.com`).

### DIFF
--- a/VIRTUAL_VACATION/vacation.pl
+++ b/VIRTUAL_VACATION/vacation.pl
@@ -51,6 +51,8 @@ our $db_name     = 'postfix';
 
 our $vacation_domain = 'autoreply.example.org';
 
+our $recipient_delimiter = '+';
+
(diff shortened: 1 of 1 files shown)


DavidGoodwin:
can you explain what your patch is doing ? 
```perl
 if ($recipient_delimiter) {
        $tmp =~ s/[\Q$recipient_delimiter\E].+$//;
 }
```

My perl is getting rusty, but I think that's just removing any recipient delimiter from the email - which doesn't make sense to me... help?

csware:
This is needed as the delimiter prevents the mail-address from being found on the database.

PR #172, opened 2018-05-04
Bootstrap Integration

abonanni:
You will find my contribution for integratiing bootstrap based template.
Interace would now be responsive.


cboltz:
I assume I need to set `$CONF['theme'] = 'bootstrap';`, right?

This indeed switches the theme, but unfortunately I get a mostly unformatted page (for example, the menu appears as unstyled bullet list). Am I doing something wrong, or do you need to fix something? ;-)

hardware:
Well done mate. Works fine with my docker image, I created a new docker tag to test your Bootstrap integration. 

```
docker pull hardware/postfixadmin:bootstrap
```

DavidGoodwin:
Looks good to me.

Should we make this the default theme?

DavidGoodwin:
@abonanni are you planning on doing any further work on feature, or should I just merge it now?

abonanni:
@DavidGoodwin : I haven't planned to work on this feature on the next weeks. So you can merge it now, and if users need some other feature i can work on it later.

DavidGoodwin:
Well, I like the look of it - but I didn't want to merge it before you'd finished .... that's all. Thank you.

cboltz:
I really like your bootstrap template, and agree with having it merged.

However, it comes with a big disadvantage - maintenance pain. You more or less copied all templates and adjusted them for bootstrap. If we do future template changes (for example adding a new field type), we have to do that twice and (more important) have to actually remember to do it twice.

Sorry if that sounds too harsh - it's not my intention. I "just" have too much experience with merging back duplicated code - if the copies had too much time to drift away from each other, it's a pain to get them in sync again. Therefore I have developed an allergic reaction on copied-then-edited code ;-)

If you have some time (doesn't need to be in the next days), I'd love to see a follow-up pull request to change the duplicated code to
- include files (for example, big parts of list.tpl might be a candidate for this)
- `{if theme=...}` blocks for small differences
- completely merged files which apply the CSS classes of both themes

The last option would of course be perfect, but I can imagine that there are situations where bootstrap and our traditional theme are too different so that you'll need one of the other options ;-)

DavidGoodwin:
* Why would we keep the older theme, given the advantages of bootstrap (mobile friendly, more responsive etc).

cboltz:
maybe because some people still like it - but yeah, good argument ;-)

I'd say let's keep it for now, but if it becomes too difficult to maintain both, dropping the old theme is an option. (Unless someone is fast enough to merge both themes, of course - keeping some if conditions is much easier ;-)

abonanni:
@cboltz : agree with you, it's difficult to maintain two versions of template and it's a problem.
When I began that integration I tried to keep the original theme and use the ``{if theme=...}`` blocks, but the HTML code of the original template and the boostrap template is so different that I prefer to have separated files.

I think a solution would be to defined fields of the form outside the template itself. For example with the symfony framework, you create form with ``FormBuilder`` component, and templates define how to render the different fields.

8ctopus:
I've just seen this and I'm going to test it. Love the bootstrap idea.

8ctopus:
I'm really impressed with the new look, it gives a modern look and feel to the app.
I would definitely consider making it the default in the next release.

matzex:
@DavidGoodwin Is it possible to include this also into the 3.2.X releases? Would be okay to have this optionally and activate it within the config file.

matzex:
@DavidGoodwin Any plans for a release with this template? 
Maybe a 3.3.X release with more of the changes/updates in the current branch. Would be very nice, thanks!

PR #181, opened 2018-05-28
Add support for difficulty setting for php_crypt

Sopsy:
Added difficulty setting for password encryption for php_crypt method. Only supports BLOWFISH, SHA256 and SHA512.

Checks the value and gives out an error if it's not valid. Empty string (default) means use default values, which means 10 for BLOWFISH and 5000 for SHA256 and SHA512.

Verified that it works fine with Dovecot and Roundcube, all others should support the "rounds=xxxx" as well.

I also added documentation for php_crypt in config.inc.php.

100% backwards compatible.


DavidGoodwin:
Thanks! Merged.

PR #199, opened 2018-08-17
Adding support for password expiration.

doktoil-makresh:
Please read README.password_expiration for more details

### DIFF
--- a/README.password_expiration
+++ b/README.password_expiration
@@ -22,7 +22,7 @@ All my tests are performed using $CONF['encrypt'] = 'md5crypt';
 
 **If you are using Roundcube's password plugin, you should also adapt the $config['password_query'] value.
 I recommend to use:
-$config['password_query'] = 'UPDATE mailbox SET password=%c, modified=now(),pw_expires_on=now() + interval 90 day, thirty=0,fourteen=0,seven=0 where username=%u';
+$config['password_query'] = 'UPDATE mailbox SET password=%c, modified=now(),pw_expires_on=now() + interval 90 day';
 of cource you may adapt to the expected expiration value
 
 All my tests are performed using $config['password_algorithm'] = 'md5-crypt';
--- a/README.password_expiration
+++ b/README.password_expiration
@@ -2,6 +2,8 @@
 This extension adds support for password expiration.
 It is designed to have expiration on users passwords. An email is sent when the password is expiring in 30 days, then 14 days, then 7 days.
 It is strongly inspired by https://abridge2devnull.com/posts/2014/09/29/dovecot-user-password-expiration-notifications-updated-4122015/, and adapted to fit with Postfix Admin & Roundcube's password plugin
+Expiration unit is day
+Expiration value for domain is set through Postfix Admin GUI
 
 *Installation
 Perform the following changes:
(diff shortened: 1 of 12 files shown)


doktoil-makresh:
Hi,

To be honest I looked for a way to deal with authentication through a file, but I spent so many time to adapt postfixadmin (I've 0 skill in PHP) that I stopped my researches. Thanks to your link, I found a way to deal with it.
My latest commit should fit security requirements. Feel free to comment

DavidGoodwin:
Hi - I like the idea, but think I'd rather ...

 * the expiration duration was based on either some value stored within the domain table itself (e.g. domain.password_expiry = <null|number-of-days>) ... this would make it configurable by a domain admin and allow different domains to have different policies (and for the value to be editable from within postfixadmin)
 * mailbox just had a pw_expiry_date and not the other boolean fields.
 * the queries in your notification script can just be scheduled to run via cron and look for mailboxes which have pw_expiry_date > (now+7 days) and pw_expiry_date < (now+8 days) or whatever the logic needs to be.

Yeah you'd send duplicate notifications if the script ran twice in a day ... but there are probably worse things (like adding unnecessary db fields!) in the world.

doktoil-makresh:
1/WiP
2/Done
3/Done

gnanet:
only a side-note, the static .my.cnf is one step away from generating it dynamically, to have the U/P inside your bash script, but not passed as a param, but by creating a temporary .cnf file, but thats also not a show-stopper overall

doktoil-makresh:
@gnanet gnanet : I understand, but when possible, I prefer to have credential in outside file, easier to protect.
@DavidGoodwin : changes applied, tested and submitted in a new pull request

DavidGoodwin:
@doktoil-makresh thank you :) I'll try and review / test the other pull request soon.

zentavr:
Hello @doktoil-makresh. I wonder how you designed the flow of setting up the password expiration value in the PostfixAdmin portal? Is the process relies only for RoundCube plugin?

doktoil-makresh:
Hello @zentavr
Postfixadmin UI is indeed used to manage expiration (enable password expiration, set the expiration value)
The suggested change in the SQL query used by RC password plugin provides a way to reset password expiration (change date + 90 days). Without this change, the password will expire even though changed recently.

So the use case is: enable password expiration in PostfixAdmin, change the SQL query in RC password plugin. If a user receives an email about password to expire soon, then he can use RC to change his password and reset password expiration

zentavr:
@doktoil-makresh actually I've enabled that `$CONF['password_expiration_enabled'] = 'YES'` feature and do not see any new fields in my PostfixAdmin which I can edit. What I can edit is the value for the domain globally - not a per-user value.
I cannot find anything related in HTML forms when I'm creating the new user or editing an existent one.

At the user portal, when the user had changed the password himself - nothing had happened with the `password_expiry` field into MySQL table, so I've submitted PR #493.

I'm seeking out if I missed smth with PostfixAdmin setup.

doktoil-makresh:
@zentavr I did this change long time ago, but I do remember the password expiration was a global value, applied to all domains managed by PostfixAdmin.
I do remember that changing user password from PostfixAdmin UI did reset the expiration to "now + 90 days" but I may be wrong

PR #200, opened 2018-08-20
Support for password expiration, managed in PostFix Admin

doktoil-makresh:
According to @DavidGoodwin's requests


DavidGoodwin:
Note : https://github.com/doktoil-makresh/postfixadmin/pull/1

PR #210, opened 2018-10-22
Change "characters" to "letters" in EN language file

JesseSteele:
A "character" includes letters and numbers. The requirement triggered by line 355 is for "letters". This created confusion in the past.


DavidGoodwin:
Thanks!

PR #216, opened 2018-12-11
Allow different port for fetchmail

SuperVirus:
Adds option to specify non-standard port for fetchmail, in case the mail server is listening on a different port.


languages/de.lang
```diff
@@ -390,6 +391,7 @@ $PALANG['pFetchmail_field_returned_text']   = 'Zurückgegebener Text';
 $PALANG['pFetchmail_desc_id']               = 'Eintrags-ID';
 $PALANG['pFetchmail_desc_mailbox']          = 'Lokales Postfach';
 $PALANG['pFetchmail_desc_src_server']       = 'Entfernter Server';
+$PALANG['pFetchmail_desc_src_service']      = 'Entfernter Service, permits you to specify a service name to connect to. You can specify a decimal port number here, if your services database lacks the required service-port assignments.';
```
cboltz: This does not match the english text - and doesn't look german ;-)

Please get the text in sync with the english text (translating it is of course ok ;-)

public/upgrade.php
```diff
@@ -1765,3 +1765,8 @@ function upgrade_1842() {
     _db_add_field('mailbox', 'password_expiry', "{DATETIME}"); // when a specific mailbox password expires
     _db_add_field('domain',  'password_expiry', 'int DEFAULT 0'); // expiry applied to mailboxes within that domain
 }
+
+function upgrade_1843() {
+   # Additional field for fetchmail to allow server with non-standard port
+    _db_add_field('fetchmail', 'src_service', "varchar(30) {UTF-8} NOT NULL DEFAULT ''", 'src_server');
```
cboltz: UTF-8 for a port name or number? Really? ;-)

Also, do you have real-world usecases where you need to specify the port, but at the same time can use a port name?

My guess would be that this is a very rare situation, and restricting the port to a port number (integer) would not be annoying to anybody ;-) and at the same time easily avoid that someone can abuse this field to inject something into the generated fetchmail.conf.

SuperVirus:
I'm fixing it later and create another PR.

PR #221, opened 2018-12-17
Allow different port for fetchmail

SuperVirus:
Adds option to specify non-standard port for fetchmail, in case the mail server is listening on a different port.


DavidGoodwin:
Thanks for the patch.

PR #224, opened 2018-12-23
Enable recaptchav2 on login screen

bofh16:
Few modification to add this support. Variables go to config.local.php. Disabled by default/if not configured.


DavidGoodwin:
I think I'm of the opinion that while this might be useful, if you want to protect your admin login page, you should really add an Apache http auth block instead?

PR #230, opened 2019-01-05
add warning on setup over http connection

8ctopus:
I added a warning when the user is setting up postfixadmin over an insecure http connection.
I actually made that mistake myself, so maybe it will be useful to other users.


DavidGoodwin:
Thanks; merged.

8ctopus:
Glad you liked it.

PR #231, opened 2019-01-08
Feature try pdo

DavidGoodwin:
pull request is for potential review/comments/discussion


DavidGoodwin:
the reasoning for this branch is so we can start supporting prepared statements, rather than having to manually escape variables with escape_string() everywhere.

It also simplifies some of the underlying database logic/calls.

PR #232, opened 2019-01-08
Fix for MySQL 8

volp1s:
The keyword ROW became reserved in MySQL 8.0.2
https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-R


DavidGoodwin:
Thanks for the patch. Merged etc.

PR #249, opened 2019-02-19
pacrypt_dovecot: Don't blacklist SCRAM-SHA-1

mtdcr:
Actually it doesn't include the username at all. Fixes commit f444de4.


mtdcr:
Do you have plans to make a new release based on master soon? Otherwise it would be nice if you could pick this patch into 3.2.

PR #255, opened 2019-03-16
Make quota levels configurable

csware:
Allow users to change the thresholds for quota levels.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -344,6 +344,10 @@ $CONF['quota'] = 'NO';
 $CONF['domain_quota'] = 'YES';
 // You can either use '1024000' or '1048576'
 $CONF['quota_multiplier'] = '1024000';
+// fill state threshold (in per cent) for medium level (displayed as orange)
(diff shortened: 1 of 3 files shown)


DavidGoodwin:
If you could add something into your code so the two new config variables ( quota_level_high_pct and quota_level_med_pct ) have default values matching what they were ( 55 and 90 ) that would be great. 

(I suspect many people will have customised their config file, and may not pick up new entries if/when they upgrade).

DavidGoodwin:
(Thanks for taking the time to contribute!)

csware:
I thought config.inc.php is the default value.

DavidGoodwin:
> I thought config.inc.php is the default value as people should use a local version.

Yes it is .... or should be. But some people only have a config.inc.php, and do not change that file when upgrading.

csware:
Please give a hint where and how to do.

DavidGoodwin:
Nevermind; the code isn't really setup to support that. Pretend I didn't ask ;-)

PR #277, opened 2019-05-31
Add option to use smtp password when sending admin emails (fixes #272)

### LINKED ISSUE #272: Add config option to set password for admin_email
We are having a problem on our server with the password reset. The `admin_email` is configured correctly and exists on the mail server. However, the mail server requires a password to send from this address. As far as I can see, Postfixadmin has no option to use a password with `admin_email`. We could maybe avoid this problem by allowing passwordless login with an IP whitelist. But it would be much simpler if we could simply configure `admin_email_password` in postfixadmin.

I can probably implement this if you agree with the idea.

Nutomic:
See #272 for the description.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -150,6 +150,12 @@ $CONF['database_tables'] = array (
 // Leave blank to send email from the logged-in Admin's Email address.
 $CONF['admin_email'] = '';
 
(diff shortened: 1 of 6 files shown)


DavidGoodwin:
Thank you

PR #281, opened 2019-06-07
Dont swallow database connection exceptions

Nutomic:
The current code catches database connection errors, and then completely ignores them, throwing an exception with no details about the original error. It cost me a lot of time until I figured out this problem. This PR just removes the exception handler, as PFA will crash in that case anyway.


DavidGoodwin:
Thanks for contributing!

PR #286, opened 2019-07-19
Use percentage for quota levels

gabrielfin:
`quota_width` goes from 0 to 120, and `percentage` from 0 to 100. I suppose the intention was to use percentages here, right?


DavidGoodwin:
Thank you :)

PR #290, opened 2019-08-23
a small mistake in user_query with quota

akhilin:
mistake in user_query with quota enabled for dovecot 1.2+
It took two days for me to find this issue.

Thank you,
Akhil


DavidGoodwin:
Thanks for the commits - in future, trying to not mix different things in a merge request is better.

PR #295, opened 2019-09-06
Update postfixadmin-cli

Corvan:
Make this script platform independent to be usable under e.g. FreeBSD, where bash is located in /usr/local/bin/bash and thus the script fails.


Corvan:
I don't know why the CI failed, that seems unrelated to the changes I made

DavidGoodwin:
thanks for the PR.

PR #298, opened 2019-09-18
remove double libmime-encwords-perl install call in virtual vacation …

leitmedium:
…howto


DavidGoodwin:
Thank you! (even for such a small change)

PR #308, opened 2019-10-19
Favicon config and ru_lang

dsamoshin:
1. Favicon configuration of the project:
  a. Fixed incorrect link to favicon from the subfolder /users/
  b. Added ability to set favicon via config
2. Completed translation of all string constants into Russian language.


DavidGoodwin:
Thank you

PR #322, opened 2020-01-07
2FA authentication

alexandregz:
2-factor authentication, only for admins.

### DIFF
--- a/config.local.php.example.2fa
+++ b/config.local.php.example.2fa
@@ -27,7 +27,7 @@ function x_struct_admin_modify($struct) {
         return $struct;
     }
 
-     $struct['x_2fa_secret'] = pacol(1, 1, 1, 'pass', '2FA_secret', '2FA_secret_desc', '', array(), 
+     $struct['x_2fa_secret'] = pacol(1, 1, 0, 'pass', '2FA_secret', '2FA_secret_desc', '', array(), 
     /*not_in_db*/ 0,
     /*dont_write_to_db*/ 0,
     /*select*/ 'x_2fa_secret'
@@ -45,7 +45,7 @@ function x_struct_admin_modify($struct) {
     $struct['x_2fa_qrcode'] = pacol(0, 0, 1, 'html', '2FA_qr_code', '2FA_qr_code_onthefly', '', array(), 
         /*not_in_db*/ 0,
         /*dont_write_to_db*/ 0,
-        'CONCAT("<img src=\"'.$urlQRCode[0].'", REPLACE(username, "@", "%40"), "'.$urlQRCode[1].'", x_2fa_secret, "'.$urlQRCode[2].$urlQRCode[3].'\">") AS x_2fa_qrcode'
+        'IF(x_2fa_secret > "", CONCAT("<img src=\"'.$urlQRCode[0].'", REPLACE(username, "@", "%40"), "'.$urlQRCode[1].'", x_2fa_secret, "'.$urlQRCode[2].$urlQRCode[3].'\">"), "") AS x_2fa_qrcode'
     );
 
     return $struct;
--- a/config.local.php.example.2fa
+++ b/config.local.php.example.2fa
@@ -1,0 +1,53 @@
+<?php
+require __DIR__ . '/vendor/autoload.php';
+
+$CONF['database_type'] = '';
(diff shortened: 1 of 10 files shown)


config.local.php.example.2fa
```diff
@@ -0,0 +1,53 @@
...
+    // all params are HTMLT-ENCODED (EMAIL, for example)
+    //https://api.qrserver.com/v1/create-qr-code/?data=otpauth%3A%2F%2Ftotp%2EMAIL%3Fsecret%3DSECRET%26issuer%3Dpostfix.admin&size=200x200&ecc=M
+
+    $urlQRCode = array('https://api.qrserver.com/v1/create-qr-code/?data=otpauth%3A%2F%2Ftotp%2F', '%3Fsecret%3D', '%26issuer%3Dpostfix.admin',
+                        '&size=200x200&ecc=M');
+
+    $struct['x_2fa_qrcode'] = pacol(0, 0, 1, 'html', '2FA_qr_code', '2FA_qr_code_onthefly', '', array(), 
+        /*not_in_db*/ 0,
+        /*dont_write_to_db*/ 0,
+        'IF(x_2fa_secret > "", CONCAT("<img src=\"'.$urlQRCode[0].'", REPLACE(username, "@", "%40"), "'.$urlQRCode[1].'", x_2fa_secret, "'.$urlQRCode[2].$urlQRCode[3].'\">"), "") AS x_2fa_qrcode'
```
s3lph, line 48: This leaks the quite a lot of information to a seemingly random third party (the QR code API provider):
- the email address of the user
- the 2FA secret of the user
- the location of the postfixadmin instance through the HTTP Refer[r]er header (depending on the referrer policy)

In addition, it allows the third party (assuming they became malicious) to deliver back arbitrary content. Think some of the [weird things](https://research.securitum.com/do-you-allow-to-load-svg-files-you-have-xss/) you can do if you return SVGs instead of PNGs.

Even if it turns out that this case is immune to such attacks, this simply makes the site admin a prime phishing target (even admins aren't 100% immune to social engineering)

And even if this were not an issue either, this introduces an implicit dependency on a third party service the user of the software is not even necessarily aware of.

s3lph: To be a bit more constructive: I'd also like to see 2FA, but maybe it'd be better to generate the QR code locally using something like phpqrcode or qrcode.js

(PS: I'm not a postfixadmin maintainer, just a user who's excited about the option to get 2FA in postfixadmin)

public/2fa.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    // we check code here
+    $fCode = safepost('fCode');
+
+    $ga = new PHPGangsta_GoogleAuthenticator();
+    if($ga->verifyCode( $row['x_2fa_secret'], $fCode, 2)) {     // 2 = 2*30sec clock tolerance
+
+        // ---- ToDo need some implementation in functions.php and login.php :-(
+        // // set cookie don't ask for 2fa next 30 days
```
DavidGoodwin, line 51: this implies I won't be asked for 2fa if I set a cookie? Doesn't this feel a little insecure? 

ideally wouldn't you want something on the user db record (e.g. a timestamp field which if not-null is the time of next checking?)

templates/2fa.tpl
```diff
@@ -0,0 +1,49 @@
...
+	</th>
+	</tr>
+	<tr>
+		<td class="label"><label>{$PALANG.2FA_form_code}:</label></td>
+		<td><input class="flat" type="text" name="fCode" /></td>
+	</tr>
+	<!-- 
+	<tr>
+		<td class="label"><label>{$PALANG.2FA_form_30days_authentication}:</label></td>
+		<td><input type="checkbox" value="1" name="f30DaysRemember"/</td>
```
DavidGoodwin, line 23: your html is missing a > here

Neustradamus:
To follow

winicius87:
2FA with TOTP should be easier to implement with one POST request that has the fields username, password, and totp.

After receiving the username and password, hide the element and show the totp input form. This way it looks like the user is sending two POST requests to the server.

Neustradamus:
@DavidGoodwin: Why have you closed it?

DavidGoodwin:
Sorry, I think I made a mistake.

PR #324, opened 2020-02-09
fetchmail dovecot delivery

smhessenauer:
Use dovecot lda delivery via the mda setting. Similar to the encrypt setting it requires the usage of a magic keyword.


DavidGoodwin:
(Thank you for taking the time to submit a PR)

smhessenauer:
Just added a short docu, hope this is sufficient. I also made the path to dovecot/delivery configurable.

DavidGoodwin:
Thank you!

PR #325, opened 2020-02-10
[Travis] Test against PHP 7.4

Slamdunk:
I keep getting this error in production, hope the test suite raises it:

`Deprecated: Function get_magic_quotes_gpc() is deprecated in ./postfixadmin-postfixadmin-3.2/functions.inc.php on line 317`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: php
 php:
   - 7.2
   - 7.3
+  - 7.4
 
 services:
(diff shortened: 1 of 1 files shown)


DavidGoodwin:
Thank you! 

As we say PHP 5.6 is the minimum version, I see no reason why we should even check for magic quotes any longer.

https://en.wikipedia.org/wiki/Magic_quotes

PR #338, opened 2020-03-05
Limit alias destinations to local domains

michaelkrieger:
Adds optional parameter to limit the alias destinations to local domains. An additional check makes sure that the domain is in the list of postfix domain names. Forwarding to a remote server can forward Spam (hurting a server's reputation) and does not work in a DMARC-world where major senders limit who can send mail on their behalf. To avoid this, an optional parameter that allows for local aliases (such as to another mailbox in the current domain or to another domain on the same server), but will prevent remote aliases (such as to gmail or yahoo) from being created. This does not impact existing aliases.


DavidGoodwin:
Alternatively you can configure your server to use SRS ? (eg. https://github.com/roehling/postsrsd )

michaelkrieger:
Your proposed comment isn’t a solution to the feature, but I do appreciate it. The PR is an important feature in a modern email world should a provider not wish to forward mail remotely. It could also be used for additional security in a company which may not want a user to forward mail to their public mail account.

I’ve used SRS using that same tool. Inevitably some Spam (or just email formatted in a way a provider doesn’t like) will always slip through and then you’re taking credit for it under your hosts main domain. SRS works around DMARC issues, but doesn’t fix that some users get huge quantities of mail of various quality and forward it to big providers which judge you for it. Those providers then delay or reject certain mail.  While we have a good track record, for the handful of folks who want to forward to gmail, we advise them to pull from gmail instead of mail forwarding it to help keep that reputation intact. The difference in our tests is measurable..

DavidGoodwin:
Is it possible someone would want to allow non-local forwards for some domains, and not all ?

michaelkrieger:
> Is it possible someone would want to allow non-local forwards for some domains, and not all ?

It's certainly possible.  Some may way to allow only certain domains (ie: a parent company, or gmail) or even disallow certain domains (ie: free mail providers).  A hook could be added on the failure of this function to allow someone to make exceptions.  I'm happy to program that into the PR if you think it's appropriate.

That said, it's probably more likely that it's an all-or-nothing (remote-vs-local) scenario, which is what this PR currently does.

DavidGoodwin:
Yes - I was thinking your changes would be better if they were moved out into some sort of hook - where the admin-type decides what rules they want.

We have other instances in the config file where a hook/callable can be specified and that's called by the code-base. It'd also allow someone to do a token SMTP/DNS check or whatever on the goto portion (if that doesn't happen already).

functions.inc.php
```diff
@@ -260,6 +260,33 @@ function check_domain($domain) {
...
+ * @return string empty if the domain is valid, otherwise string with the errormessage
+ */
+function check_localaliasonly($domain) {
+    // If emailcheck_localaliasonly is set to 'YES', disallow aliases to remote servers (but allow aliases on this server)
+    if (Config::bool('emailcheck_localaliasonly')) {
+        // get the domain part of the e-mail
+        list(/*NULL*/, $domain) = explode('@', $domain);
+        
+        // get all domains managed on this system by postfixadmin
+        $domains = list_domains();
```
cboltz, line 275: Would it make sense to limit this to domains visible to the currently logged in admin? 

`list_domains_for_admin($username)` does that.

Background: By allowing all domains, this could leak information if a domain is hosted in the same PostfixAdmin instance (by trying various alias targets), even if an admin doesn't have permissions for this domain.

Sidenote: Since this function is only used by AliasHandler, I'd recommend to have it there instead of making functions.inc.php even bigger.

michaelkrieger: I think that could easily be configurable even:
`$domains = ( (Config::bool(‘emailcheck_localaliasownedonly’)) ? list_domains_for_admin($username) : list_domains() );` 

My end goal was more to prevent forwards to hotmail/gmail (in favour of pulling such data instead of pushing it via SMTP), but your point that this could leak data is certainly valid.

I’d think that’s appropriate to use list_domains_for_admin(), so I’ll agree with you.

svenseeberg:
I'll first add to this conversation, because this is already going somewhat into the directions that are important for me.

We are currently also having issues with domain postmaster creating aliases with lots of external goto addresses. While the DMARC problem can and should be solved with postsrsd, having too many goto addresses still poses a problem because it amplifies spam. Our spam filter is catching some, but not all spam mails. Sometimes still dozens of mails still get forwarded, which can really hurt the mail server reputation. Therefore I think it should be possible to impose some sort of limitations for aliases. Currently, I'm thinking about adding a setting that simply limits the number of gotos for one alias. I think a limit of 10 is pretty reasonable. In all cases where there were more recipients, the alias function was basically a mailing list replacement / abuse.

And my thoughts about limiting gotso to local domains: I don't think this is necessary. This needs to be solved by rewriting the sender address with postsrsd.

functions.inc.php
rpv-tomsk: >This could leak information if a domain is hosted in the same PostfixAdmin instance 

I don't think such leak is really critical, because these details could be easily explored from DNS.

michaelkrieger:
This hasn't moved since 2020 and is an increasingly important feature (to not get flagged as a spammer by major providers where mail could be forwarded) with minimal additional code (only a few lines).  The patch applies cleanly to the current codebase.  Can we please see this merged @DavidGoodwin ?

PR #343, opened 2020-03-25
Initial commit for Autoconfig

jackdeguest:
Added Autodiscovery configuration interface to Postfix Admin version 3.3 (?).
Files added are:
- AUTOCONFIG folder containing elements to be moved, including sql scripts for the 3 database drivers supported.
- en.lang, fr.lang, ja.lang updated
- config.inc.php updated to contain certain preferences pertaining to payload signature
- functions.inc.php adding 3 new functions: db_begin, db_commit and db_rollback
An INSTALL.md is added giving instruction including on how to set up Apache Virtual Host and DNS zone. Maybe need to add information for Nginx?
Autoconfig uses foreign key to work properly.
This addition has been tested thoroughly by me but some bugs may have slipped my attention.


AUTOCONFIG/AutoconfigHandler.php
```diff
@@ -0,0 +1,1284 @@
...
+            'account_type'				=> pacol( 1, 		1,      1,      'text', 'Autoconfig_account_type'				, ''                                , '' ),
+            'email'						=> pacol( 1, 		1,      1,      'text', 'Autoconfig_email'				, ''                                , '' ),
+            'ssl_enabled'				=> pacol( 1, 		1,      1,      'boolean', 'Autoconfig_ssl'				, ''                                , '' ),
+            // Will be empty obviously unless the user enters it in the form
+            'password'					=> pacol( 1, 		1,      1,      'text', 'Autoconfig_password'				, ''                                , '' ),
+            // Used for payload_description
+            'description'				=> pacol( 1, 		1,      1,      'text', 'Autoconfig_description'				, ''                                , '' ),
+            'organisation'				=> pacol( 1, 		1,      1,      'text', 'Autoconfig_organisation'				, ''                                , '' ),
+            // regular account, or Microsoft Exchange
+            'type'						=> pacol( 1, 		1,      1,      'text', 'Autoconfig_type'				, ''                                , '' ),
```
DavidGoodwin: you've defined the 'type' column twice....

AUTOCONFIG/AutoconfigHandler.php
```diff
@@ -0,0 +1,1284 @@
...
+        // This is a super admin, so he/she has access to all configs
+        if ( authentication_has_role( 'global-admin' ) ) {
+            // $sql = "SELECT DISTINCT ad.config_id FROM $table_autoconfig_domains ad LEFT JOIN $table_domain d ON ad.domain = d.domain WHERE d.domain != 'ALL AND d.active IS TRUE'";
+            // global admin has access to all config
+            $sql = "SELECT c.config_id FROM $table_autoconfig c";
+        }
+        // This is a per-domain admin, so we use the table domain_admis to cross check which configuration he/she has access
+        elseif ( authentication_has_role( 'admin' ) ) {
+            $E_username = escape_string( $user );
+            $sql = "SELECT DISTINCT ad.config_id FROM $table_domain d LEFT JOIN $table_autoconfig_domains ad ON ad.domain = d.domain WHERE d.active IS TUE AND d.username='$E_username'";
```
DavidGoodwin: is this TUE a typo?

Can you rewrite it so there's no chance of $sql being undefined. 
E.g. default $sql to whatever the 'admin' user has ?

jackdeguest: Good catch. Yes, this should have been `TRUE`

AUTOCONFIG/AutoconfigHandler.php
jackdeguest: Actually, I did not really understand the purpose of this `initStruct()`, and am not using it. Yes, the column appeared twice and this is a mistake. `type` is used as a property of the overall `autoconfig` and also as a property of the hosts `autoconfig`. My bad.

DavidGoodwin:
> Thank you kindly for catching and correcting my typos !

Most of them are from running 'composer psalm' and 'composer format' on the code. 

I haven't done much yet.

HLFH:
@jackdeguest @DavidGoodwin Any updates about status? I was thinking it was a good idea to try to integrate Autoconfig, Mobileconfig, potentially Autodiscover, and other autodiscovery helper mechanisms into Postfixadmin, so the email user only needs to provide an email address and a password.

AUTOCONFIG/AutoconfigHandler.php
knofte: ( and probably 'table domain_admins' in the comment :+1:

ETE-Design:
@jackdeguest Would be great with information for Nginx also.

PR #355, opened 2020-05-15
Make pInvalidDomainDNS more forgiving

rootwyrm:
Make the pInvalidDomainDNS check more forgiving to accommodate newly created domains which also do not have an A record for the domain (which is generally not a best practice) or MX configured yet.
This adds a check for an `NS` record; if the NS exists, then the domain is valid and can be assumed to have just not been configured to receive email yet.

### DIFF
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -248,6 +248,8 @@ function check_domain($domain) {
                 $retval = '';
             } elseif (checkdnsrr($domain, 'MX')) {
                 $retval = '';
+            } elseif (checkdnsrr($domain, 'NS')) {
+                $retval = '';
             } else {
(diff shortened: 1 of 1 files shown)


rootwyrm:
One change that I mulled over before sending the PR is additionally logging a message warning that the domain cannot be used to send/receive email when the `NS` check is hit, e.g.

```
} elseif (checkdnsrr($domain, 'NS')) {
    error_log("DNS is not correctly configured for $domain to send or receive email");
    retval = '';
} else {
```

DavidGoodwin:
I suspect this fixes an edge case you encounter; I don't know if many others encounter the same problem! :)

Anyway - merged. I'll add the error_log in.

DavidGoodwin:
Thanks for taking the time to contribute!

PR #361, opened 2020-06-15
fix vacation.pl resolver

ymarkus:
in commit [86852b3dbdbc8b9a3923d5186d2f32f630328ccf](https://github.com/postfixadmin/postfixadmin/commit/86852b3dbdbc8b9a3923d5186d2f32f630328ccf) there is a [comment](https://github.com/postfixadmin/postfixadmin/commit/86852b3dbdbc8b9a3923d5186d2f32f630328ccf#r39604078) stating that "`$email` is the original email recipient.
The MX should be found from `$orig_from`."

But this leads to vacation.pl trying to send mail via the original mail senders server which is wrong in my eyes and fails for basically every server.

After applying this patch, it actually uses the server of the original recipient. 
Please check again, @Erwane @DavidGoodwin. 

Here the lines from my log indicating something is wrong:
```
DEBUG> /usr/lib/postfixadmin/vacation.pl:464 main::send_vacation_email - Found MX record <gmail-smtp-in.l.google.com> for user <user@example.com>!
ERROR> /usr/lib/postfixadmin/vacation.pl:514 main::finally {...}  - Failed to send vacation response to user@gmail.com from ym@ymarkus.dev subject Out of Office: unable to establish SMTP connection to gmail-smtp-in.l.google.com port 25
```
Google is NOT! the correct mx record for my domain (which I replaced here with example.com)

Here after the patch:
```
DEBUG> /usr/lib/postfixadmin/vacation.pl:464 main::send_vacation_email - Found MX record <mail.example.com> for user <user@example.com>!
DEBUG> /usr/lib/postfixadmin/vacation.pl:516 main::finally {...}  - Vacation response sent to user@gmail.com from user@example.com subject Out of Office sent
```
Now the MX record is correct!

### DIFF
--- a/VIRTUAL_VACATION/vacation.pl
+++ b/VIRTUAL_VACATION/vacation.pl
@@ -454,7 +454,7 @@ sub send_vacation_email {
         my $to = $orig_from;
 
         # part of the username in the email && part of the domain in the email
-        my ($email_username_part, $email_domain_part) = split(/@/, $orig_from);
+        my ($email_username_part, $email_domain_part) = split(/@/, $email);
 
         my $resolver  = Net::DNS::Resolver->new;
         my @mx   = mx($resolver, $email_domain_part);


DavidGoodwin:
Thanks; I'm not actively using vacation.pl at the moment, so my testing ability is limited.

Erwane:
Patch looks good. Thanks.

ymarkus:
Thanks!

PR #375, opened 2020-08-07
Disallow alias that points to itself (fixes #358)

### LINKED ISSUE #358: Do not allow alias that points to itself when there is no matching mailbox
I've been working with PostfixAdmin for a while and I've encountered an issue which I think is a specific logic issue in the interaction between PostfixAdmin's database, Postfix and final mail delivery (Dovecot).

For better understanding, these are some simplified steps that hopefully reproduce the problem:

1. Have working Postfix, PostfixAdmin and Dovecot, with domains, mailboxes and aliases.
2. Configure Postfix to send email via LMTP to dovecot. Dovecot accesses the same PostfixAdmin database for user/mailbox lookups. Configure Postfix to reject recipient addresses that do not exist in the database. (The latter two things can be easily achieved with the docs included in this project)
3. Create a catch-all alias for a domain (say, example.com) that points to mailbox@example.com
4. Create the mailbox "mailbox@example.com" in PostfixAdmin
5. Create an alias "loopback@example.com" that points to "loopback@example.com"
6. DO NOT create a mailbox with the same name.
7. Send email to postfix, using "loopback@example.com" as recipient address
8. Watch how Postfix accepts the email, aliasing the recipient to "loopback@example.com" and forwards the email with "loopback@example.com" as final, aliased recipient (incorrect)
9. Watch how dovecot fails to deliver the mail, as there is no mailbox at that address.

The setup above works just fine when the "loopback@example.com" alias does not exist. I think the issue is that when Postfix verifies the recipient address in the database, it first looks up the alias, which has a match in the database. However, the given alias target has no matching mailbox in the database.

But, since there is a catch all, if Postfix continues the lookup it will find the catch all and thus finds a mailbox to deliver the mail to. Normally, Postfix would alias the recipient address and resolve to the final mailbox address, which is then passed through to dovecot.

However, in this case where the alias with nonexistent mailbox points to itself, Postfix misinterprets the data and thinks that "loopback@example.com" is the final, aliased address of the mailbox. Thus postfix tries to deliver email using that as the recipient address, which will not work with Dovecot, because Dovecot cannot find a mailbox under "loopback@example.com" (which is correct). This causes late bounces of the mail, which would normally be prevented. But in this scenario Postfix believes that a recipient address exists even though it does not.

I think the solution to this is to prevent such "points to itself" aliases in the database, unless there is a mailbox under that name - in which case the entire problem is gone. It would be nice if PostfixAdmin would automatically forbid such aliases. Right now I need to educate my domain admins to "not do this".

PS: I'm running PostfixAdmin v3.2.4, Postfix v3.4.10, Dovecot v2.3.10.1

PPS: Just for clarification: This issue does not happen, in the same setup, when using any other type of alias combination, e.g "alias1@example.com" -> "alias2@example.com" (nonexistent) -> catchall works just fine. The only alias that fails is a "points to itself" alias without any matching mailbox.

PPPS: I just noticed that the same late bounce also happens when using the Postfix virtual delivery agent, so it doesn't really matter if Postfix relays mail to dovecot or delivers it itself.

GermanCoding:
So, hopefully this is better (followup to c3d5b26740d2461f86a18f718bffc7db26cf49c5).

The basic issue is described in #358: Aliases that "loop-back" should only be used if there is a corresponding mailbox at that address. Otherwise Postfix might fail to correctly alias the recipient address, causing mail delivery errors and late bounces (backscatter).

The new behavior prevents the user from creating such invalid aliases. The MailboxHandler is still able to create "loop-back" aliases, because those are required when adding a new mailbox. The behavior is not configurable, because I couldn't find a single setup where loopback aliases where actually serving any purpose and without causing severe trouble.

The error message is now translatable (this wasn't the case in the original commit).

The original commit was tested (in production) on my own instances. This new commit received minimal testing, but I briefly verified that all still seemed ok.

### DIFF
--- a/languages/bg.lang
+++ b/languages/bg.lang
@@ -127,6 +127,7 @@ $PALANG['alias_updated'] = 'The alias %s has been updated!'; # XXX
 $PALANG['pCreate_alias_catchall_text'] = 'За да създадете catch-all използвайте "*" за alias. За пренасочване на домейн към домейн използвайте "*@domain.tld" в полето Към.'; # XXX check/beautify - was split in two lines before
 $PALANG['mailbox_alias_cant_be_deleted'] = 'This alias belongs to a mailbox and can\'t be deleted!'; # XXX
 $PALANG['protected_alias_cant_be_deleted'] = 'The alias %s is protected and can only be deleted by a superadmin'; # XXX
+$PALANG['alias_points_to_itself'] = 'Alias may not point to itself'; # XXX
 
 $PALANG['pEdit_alias_welcome'] = 'Редактиране на alias за вашия домейн.'; # XXX Text change to: 'Edit forwarding settings'
 $PALANG['pEdit_alias_help'] = 'Един запис на ред.'; # XXX Text change to: 'Accepts multiple targets, one entry per line.'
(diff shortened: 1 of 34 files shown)


DavidGoodwin:
Hi - thanks - I was planning to write a trivial unit test to cover this before merging it ....

PR #391, opened 2020-09-26
Feature refactor login etc

DavidGoodwin:
Possible improvements to the code, probably more to do.

Changes default theme to be bootstrap.

Starts adding some PHP 7.0 style type hints for argument or return values ( see e.g. https://www.php.net/manual/en/functions.returning-values.php ).


config.inc.php
```diff
@@ -692,6 +692,8 @@ function x_send_sms_clickatell($to, $message) {
 //More details in README.password_expiration
 $CONF['password_expiration'] = 'YES';
 
+$CONF['version'] = '3.3-dev';
```
cboltz, line 695: Moving `$version` to config.inc.php doesn't look like a good idea to me. It's part of the code, not the configuration, so it shouldn't be configurable ;-) and besides that, we'll run into problems if someone still doesn't use a config.local.php.

model/PFAHandler.php
```diff
@@ -526,21 +526,21 @@ protected function setmore(array $values) {
...
+     * save $this->values to the database
      *
      * converts values based on $this->struct[*][type] (boolean, password encryption)
      *
      * calls $this->storemore() where additional things can be done
      * @return bool - true if all values were stored in the database, otherwise false
      *     error messages (if any) are stored in $this->errormsg
      */
-    public function store() {
+    public function save() : bool {
```
cboltz, line 537: What's the reason for renaming `store()`, `beforestore()` and `storemore()`?

The problem I see is that external users of PostfixAdmin will have to adjust their code / additional classes, without giving any benefit. (In case you wonder - I know at least one case of PostfixAdmin usage with additional classes that won't fit for upstream because they are totally unrelated to mail handling.)

If you really want to rename these functions, please keep compatibility calls to their old names to avoid breaking those additional classes.

config.inc.php
DavidGoodwin: It's part of 'config.inc.php' which contains the default config  ...vs the local override (config.local.php) was kind of my thinking.

model/PFAHandler.php
DavidGoodwin: preSave(), save() and postSave() seemed more descriptive and follow a more widespread naming convention.

I'll add redirection functions in to avoid issues.

PR #409, opened 2021-01-06
tutorial link updated

cool-nicq:
bliki.rimuhosting.com is not reachable.
www.linuxbabe.com has complete tutorial of installation postfix, dovecot and postfix admin


DavidGoodwin:
Thanks for taking time to contribute.

Issue #408, opened 2021-01-03
Support for dovecot mail-crypt-plugin?


BotoX:
I've recently setup a mostly standard virtual mail server with Postfix + Dovecot.
However I've decided to go with per-user encrypted maildir, using the users login passphrase.
For more info see: "Encrypted user key"
https://doc.dovecot.org/configuration_manual/mail_crypt_plugin/

When adding a new mailbox one has to pass the user + password to the following command in order to generate the encryption keys:
`doveadm -o plugin/mail_crypt_private_password="PASSWORD" mailbox cryptokey generate -u "user@example.org" -U`

Likewise when the user changes their password, the encryption keys have to be updated:
`doveadm mailbox cryptokey password -u "user@example.org" -o "OLD_PASSWORD" -n "NEW_PASSWORD"`
The old password is required to decrypt the key, which is then encrypted with the new password.

I've patched the latest stable release of postfixadmin (3.2.4) and tested it a bit, here are my changes:
https://github.com/BotoX/postfixadmin/commit/d88ed67a1aa2e3609e3186d77e219687eadc03f4

And here it is for the master branch:
https://github.com/BotoX/postfixadmin/commit/c6ae3ea2f3921c1a682315d1205bec42aaa32035

Should be flexible enough for similar setups.

Changing mailbox passwords without entering the old password (in the admin panel) is blocked by this patch, as this makes no sense.
If the user loses their mailbox password then all of their data is gone.
Changing the password in SQL without updating the encryption keys only makes dovecot scream in horror and crash.

This mail server setup is not fully in "production" yet (me and a few friends) so I can't tell how stable this dovecot feature is.

Inspired by: https://gist.github.com/yajrendrag/203b0172fee96a8b002a026362d27bf2#additional-postfixadminroundcube-setup-to-support-ispmail-optional-encryption
This also adds a new x_crypt option to the database to decide about encryption per mailbox, unlike my patch.

DavidGoodwin:
Thanks for this - I've merged it into master.

DavidGoodwin:
Hmm .... I had to comment out some of the code in Login - 

https://github.com/postfixadmin/postfixadmin/blob/70bf59edc4920ae7825857cbe76482d6a66722fc/model/Login.php#L125

as e.g. $this->id isn't defined.

What command should be being run here after someone's changed their password?

BotoX:
I guess I should've tested the changes on master too instead of just blindly copy pasting stuff...
[...]

PR #421, opened 2021-01-14
Suport for dovecot mail-crypt-plugin via new mailbox_postpassword_script hook

DavidGoodwin:
Uses doveadm mailbox cryptokey on create user / password change to adjust user encryption key.
https://doc.dovecot.org/configuration_manual/mail_crypt_plugin/

see also : https://github.com/postfixadmin/postfixadmin/issues/408


ean365:
@BotoX and @DavidGoodwin, please see my recommended change to this in #441, with regard to putting plaintext password on doveadm command line.  (I probably should have put it here.)

PR #424, opened 2021-01-17
List of addresses needs a Line Feed

gianlucagiacometti:
Edit field does display a list of aliases in a single line in the edit form. Adding a Line Feed character fixes this problem.


DavidGoodwin:
Thanks for contributing!

DavidGoodwin:
merged into postfixadmin 3.3 as well...

PR #425, opened 2021-01-17
Add a note for new fields

gianlucagiacometti:
It took me some time to understand that adding fields through the struct hook does not mean that they are automatically displayed in the corresponding virtual list. I think that adding a note here would make life easier to everybody.


DavidGoodwin:
Thank you; merged.

PR #457, opened 2021-02-19
Change Login changePassword, fixes #456

### LINKED ISSUE #456: SQL error when user tries to change password
On Postfixadmin 3.3.7, a user trying to change the mailbox password causes the following error message:
```
Invalid query: SQLSTATE[42000]: Syntax error or access violation: 1103 Incorrect table name '' caused by UPDATE ``mailbox`` SET  password = :password  WHERE username = :where, referer: https://example.com/postfixadmin/users/password.php
```

It seems that `$this->key_table` in 
https://github.com/postfixadmin/postfixadmin/blob/1a9d9847f5c5371b3ba756e9f21e3d694437c976/model/Login.php#L113 contains the string "\`mailbox\`", including backticks. I have not yet evaluated where the wrong string comes from. Replacing the parameter value with a literal `mailbox` string fixes the issue (which is currently my workaround in place).

I can see that the Login.php changed heavily since the 3.3.7 tag, but the line causing the error seems to be the same.

svenseeberg:
When the function `changePassword` is called, it passes the table name to `db_update()` which already was treated by the `table_by_key()` function. However, `db_update()` also calls the `table_by_key()` function for the table name argument. This results in double backticks and causes PHP errors for users that want to change their password.

Signed-off-by: Sven Seeberg <mail@sven-seeberg.de>

### DIFF
--- a/model/Login.php
+++ b/model/Login.php
@@ -110,7 +110,7 @@ class Login {
             'password' => pacrypt($new_password),
         );
 
-        $result = db_update($this->key_table, 'username', $username, $set);
+        $result = db_update($this->table, 'username', $username, $set);
 
         if ($result != 1) {
             db_log($domain, 'edit_password', "FAILURE: " . $username);


DavidGoodwin (approved):
Thanks ! I saw a similar problem elsewhere which was triggered by a unit test.

PR #482, opened 2021-05-07
Don't reply to known auto generated mails

csware:
Adds support for Amazon notifications, add support for "X-Auto-Response-Suppress: all", and adds some commonly used noreply addresses.


DavidGoodwin:
Thank you (again!)

PR #483, opened 2021-05-07
Be a good citizen and include 'Auto-Submitted: auto-replied' in auto replies

csware:
This is a SHOULD "requirement" in RFC3834.


DavidGoodwin:
Thank you!

Issue #337, opened 2020-03-03
Mojibake in message sent from vacation.pl


jackdeguest:
When adding a vacation text in non us-ascii such as Japanese, or French, then this vacation script will send a mail with mojibake.
The software does properly transcode mail from utf8 in database into perl internal extended utf8 version, but does not encode it back to standard utf8 before creating the mail.
Version used is Postfixadmin 3.2 and vacation version 4.2
I have attached a patch that solves this problem. Confirmed with proper testing.
[vacation.patch.txt](https://github.com/postfixadmin/postfixadmin/files/4280388/vacation.patch.txt)

DavidGoodwin:
Thank you!

PR #484, opened 2021-05-07
UTF8 is no real UTF8 in MySQL, one needs to use utf8mb4

csware:
This allows to use emojis in vacation mails and allow characters in names.

The usage of UTF8mb4 should be the default or at least configurable. utf8mb4 was introuced in MySQL 5.5.3 IIRC.

### DIFF
--- a/VIRTUAL_VACATION/vacation.pl
+++ b/VIRTUAL_VACATION/vacation.pl
@@ -208,7 +208,7 @@ if (!$dbh) {
 
 my $db_true; # MySQL and PgSQL use different values for TRUE, and unicode support...
 if ($db_type eq 'mysql') {
-    $dbh->do('SET CHARACTER SET utf8;');
+    $dbh->do('SET CHARACTER SET utf8mb4;');
     $db_true = '1';
 } else { # Pg
     $dbh->do("SET CLIENT_ENCODING TO 'UTF8'");
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -1626,9 +1626,9 @@ function db_connection_string()
         $database_name = Config::read_string('database_name');
 
         if ($socket) {
-            $dsn = "mysql:unix_socket={$socket};dbname={$database_name};charset=UTF8";
+            $dsn = "mysql:unix_socket={$socket};dbname={$database_name};charset=utf8mb4";
         } else {
-            $dsn = "mysql:host={$CONF['database_host']};dbname={$database_name};charset=UTF8";
+            $dsn = "mysql:host={$CONF['database_host']};dbname={$database_name};charset=utf8mb4";
         }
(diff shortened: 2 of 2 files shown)


csware:
Also relates to #337.

csware:
Of course at least the mailbox.name and vacation.subject and vatation.body columns need to be utf8mb4 (i use `utf8mb4_bin` as no caseinsensitive comparisons are needed).

DavidGoodwin:
current vacation table looks like : 

```
CREATE TABLE `vacation` (
  `email` varchar(255) NOT NULL,
  `subject` varchar(255) NOT NULL,
  `body` text NOT NULL,
  `activefrom` timestamp NOT NULL DEFAULT '2000-01-01 00:00:00',
  `activeuntil` timestamp NOT NULL DEFAULT '2038-01-18 00:00:00',
  `cache` text NOT NULL,
  `domain` varchar(255) NOT NULL,
  `interval_time` int(11) NOT NULL DEFAULT 0,
  `created` datetime NOT NULL DEFAULT '2000-01-01 00:00:00',
  `modified` timestamp NOT NULL DEFAULT current_timestamp(),
  `active` tinyint(1) NOT NULL DEFAULT 1,
  PRIMARY KEY (`email`),
  KEY `email` (`email`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1 COMMENT='Postfix Admin - Virtual Vacation'
```

DavidGoodwin:
alter table vacation change body body text charset utf8mb4 not null;

etc.

functions.inc.php
```diff
@@ -1699,8 +1699,8 @@ function db_connect()
 
             $options[PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT] = (bool)$verify;
         }
-        $queries[] = 'SET CHARACTER SET utf8';
-        $queries[] = "SET COLLATION_CONNECTION='utf8_general_ci'";
+        $queries[] = 'SET CHARACTER SET utf8mb4';
```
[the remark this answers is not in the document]
csware: See second commit

VIRTUAL_VACATION/vacation.pl
```diff
@@ -208,7 +208,7 @@
 
 my $db_true; # MySQL and PgSQL use different values for TRUE, and unicode support...
 if ($db_type eq 'mysql') {
-    $dbh->do('SET CHARACTER SET utf8;');
```
DavidGoodwin, line 211: Should perl be using SET NAMES ... here ?

csware: Possibly, works this way for me.

csware:
This was quick! I had expected that mergig this would take longer ;)

Have a nice weekend!

PR #491, opened 2021-05-09
Feature improve pacrypt (reduce reliance on dovecot pw)

DavidGoodwin:
Attempt to re-implement many of the 'dovecot:xxxxx' $CONF['encrypt'] variants in a PHP based implementation (relying on e.g. the 'hash' module or probably 'argon2' if you want argon2i/argon2id).

Need to find out whether the {PREFIX} on hashes causes problems for non-dovecot installations.

Need to consider whether support for other base64 encoded / hex encoded variants are needed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "psalm": "@php ./vendor/bin/psalm --no-cache --show-info=false "
     },
     "require": {
-        "php": ">=7.0"
+        "php": ">=7.2"
     },
     "require-dev": {
         "ext-mysqli":  "*",
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -1288,65 +1288,12 @@ function _php_crypt_random_string($characters, $length)
  * @param string $pw_db optional encrypted password
  * @return string encrypted password - if this matches $pw_db then the original password is $pw.
  */
-function pacrypt($pw, $pw_db="")
+function pacrypt($pw, $pw_db = "")
 {
     global $CONF;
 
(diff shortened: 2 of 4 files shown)


DavidGoodwin:
probably need to add support for looking at the start of the hash ($pw_db) and see if it matches { ... } but not {ALGO} -> and if so, service using a different alogirthm (perhaps also transparently upgrading the user to the new/currenct algorithm?)

darix:
why not split this out into a php library so that other mail related projects could use it? @cboltz

DavidGoodwin:
@darix Laziness. :-) 

It's also not a particularly elegant PHP library (no namespace, all in one file, probably a lot of missing/incorrect assumptions and features etc) which would ideally all change before it was published as a library.

darix:
> @darix Laziness. :-)
> 
> It's also not a particularly elegant PHP library (no namespace, all in one file, probably a lot of missing/incorrect assumptions and features etc) which would ideally all change before it was published as a library.

we would appreciate this step. then we could use it in https://github.com/openSUSE/ap4rc

darix:
JFYI a dovecot upstream dev just suggested instead of setting `{SHA512-CRYPT}`, that you want `{CRYPT}`

DavidGoodwin:
I'll try and extract it and publish it via packagist soon (unless someone beats me to it).

DavidGoodwin:
I've created github.com/postfixadmin/pacrypt - it's a bit empty, but I'll try and put the code into there soon.

PR #493, opened 2021-05-12
Adding an ability to update password_expiry field when user changes its own password from /users/ portal

### LINKED ISSUE #492: The password_expiry field get never updated when user changes his password (v3.3.8)
When the user comes to `/users/password.php` and tries to change his password - the password gets changed, but the `password_expiry` column is not updated.

Also, there is no possibility to change that field over the admin portal as well - only manually by jumping inside MySQL console.

zentavr:
Fixes #492


DavidGoodwin:
Thank you!

Issue #441, opened 2021-01-25
Post-password not running


NicoloLazzaroni:
Hello,

I am sorry for disturbing you.
I have a problem with the postpassword running script.

I finally implemented by user-password mail-crypt encryption. I am using the script you provided in addons to change the password of the user.

I've inserted it in the config as: `$CONF['mailbox_postpassword_script'] = 'sudo /etc/postfix/postpassword.sh';`

If I run it in the bash manually as `sudo /etc/postfix/postpassword.sh 'EMAIL' 'DOMAIN' "" 'PASSWORD'` it does work creating the keys, however if I run it in postfix admin by creating a new account I get the error:
`"The mailbox postpassword script failed, check the error log for details!"`.

I've checked the Apache2 logs and I just get `"Running sudo /etc/postfix/postpassword.sh 'EMAIL' 'DOMAIN' "" 'PASSWORD' yielded return value=1, first line of output=, referer: https://CENSORED/edit.php?table=mailbox&domain=CENSORED"`, but nothing happens and the get does not get created.

Is there anything I should change? I've also tried running it with `bash`.

Thank you very much.
Best Regards

DavidGoodwin:
Hi - that functionality should only be in master at the moment - and I don't suspect it's been tested very well yet.

To try and help figure out what's going on, can you ....

 1. Create a 'phpinfo.php' style page (containing: `<?php phpinfo(); ?>`) then view it in your browser and see what value is set for error_log ...
 2. Look at that error_log file and see if there's anything helpful shown.

I've changed model/Login.php in 84ad0ebec79a01126dc821646b2e074245a7d076 which might help it log something better.

NicoloLazzaroni:
Hello,

Thank you for your help.
I solved the running issue by using visudo and giving www-data the permission to sudo the command. I forgot to earlier.

However, I found a problem with the interface. I am debugging everything, but I saw that when adding a new user the option to add an additional recover email address disappeard, and when editing a user I can't edit the password and the additional email address. I had to edit them in MySQL.

May you add these options back? As a postadmin I can tell you they where very very useful.

Thank you.
Best Regards

NicoloLazzaroni:
Hello,

For the recover problem, I think I found one of the issues:

I changed `$CONF['forgotten_admin_password_reset'] = false` to
`$CONF['forgotten_admin_password_reset'] = true;` to permit admins password recovers, however this disabled everything.

[...]

PR #505, opened 2021-06-04
Improve dovecot mail-crypt postpassword script security

BotoX:
pipe password(s) through STDIN to postpassword script instead of passing it as arguments on cmdline.

Suggested from: https://github.com/postfixadmin/postfixadmin/issues/441#issuecomment-774736944

Also for those wanting to use this on the stable (3.3) branch, I've backported the feature here: https://github.com/BotoX/postfixadmin/tree/postfixadmin_3.3-crypt

### DIFF
--- a/ADDITIONS/postfixadmin-mailbox-postpassword.sh
+++ b/ADDITIONS/postfixadmin-mailbox-postpassword.sh
@@ -3,11 +3,14 @@
 # Example script for dovecot mail-crypt-plugin
 # https://doc.dovecot.org/configuration_manual/mail_crypt_plugin/
 
+IFS= read -r -d $'\0' OLD_PASSWORD
+IFS= read -r -d $'\0' NEW_PASSWORD
+
 # New user
-if [ -z "$3" ]; then
-    doveadm -o plugin/mail_crypt_private_password="$4" mailbox cryptokey generate -u "$1" -U
-    exit 0
+if [ -z "$OLD_PASSWORD" ]; then
+    OLD_PASSWORD="$(openssl rand -hex 16)"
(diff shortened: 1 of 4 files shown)


DavidGoodwin:
I understand this PR will remove the password from being as visible, but it's still going to be visible through these processes isn't it?

 * `doveadm mailbox cryptokey password -u "$1" -o "$OLD_PASSWORD" -n "$NEW_PASSWORD"`
 * `doveadm -o plugin/mail_crypt_private_password="$NEW_PASSWORD" mailbox cryptokey generate -u "$1" -U`

or is doveadm rewriting it's /proc entry to hide it?

BotoX:
Ouch, that was stupid of me to not think of this.
Nice catch.

I'll update the example bash script tomorrow to use:
```
-O Ask for old password
-N Ask for new password
```

And pipe the password there as well.

BotoX:
fixed the issue :+1: 

It's a little weird in the sense that I generate a random password for new accounts now to first generate the crypto keys.
Then I change the password right after with the actual one.
Also the brought up issue from https://github.com/postfixadmin/postfixadmin/issues/441#issuecomment-774736944 with having to use `-O ""` instead of just `-O`.

Works fine in my environment, which looks like this:
```
    mail_crypt_curve = secp521r1
    mail_crypt_save_version = 2
    mail_crypt_require_encrypted_user_key = yes

password_query = SELECT username AS user, password, '%w' AS userdb_mail_crypt_private_password FROM mailbox WHERE username = '%u' AND active = '1';
```

DavidGoodwin:
Thanks!

PR #508, opened 2021-06-16
Remove unnecessary <br>'s

gabrielfin:
This removes some unnecessary spacing added by `flash_error.tpl` in all pages.

Before:
before

After:
after


DavidGoodwin:
Thank you!

PR #509, opened 2021-06-16
Navbar is supposed to be fixed

gabrielfin:
The navbar had the class `fixed-top`, but it doesn't exist. It used to be `navbar-fixed-top` which indeed makes the navbar fixed, but that class was replaced in 897e57d4036bffb9779d71494e710d847297ea6a. Why was it replaced? Was it unintentional?
Also because of this, a lot of extra space is added after the navbar.

Before:
before

Now:
after

Maybe some of that space was intentional, in which case it should be added back with CSS. Let me know if I should do that.

### DIFF
--- a/templates/menu.tpl
+++ b/templates/menu.tpl
@@ -8,7 +8,7 @@
 {/strip}
 
 {strip}
-    <nav class="navbar navbar-default fixed-top">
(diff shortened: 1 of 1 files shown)


DavidGoodwin:
I've no idea why the class changed; thanks for noticing and submitting a fix !

PR #510, opened 2021-06-16
Disable password autocomplete

gabrielfin:
Passwords should not be autocompleted (by browser or password manager) in fields that ask for a new password.

I'm not sure about the change in `editform.tpl`, it seems a bit "hacky". Maybe the right way would probably be to add a new attribute to the model columns.


DavidGoodwin:
Thanks!

PR #512, opened 2021-07-05
Redirect to listview after editing

gabrielfin:
This PR redirects the user to the list view after editing an item, which is the same behaviour than after creating an item.

Was there a specific reason for the previous behaviour?


DavidGoodwin:
I don't remember .... Thanks for taking the time to contribute and improve PostfixAdmin.

PR #513, opened 2021-07-05
Allow super admins to view all log entries

gabrielfin:
This PR adds an empty option (and it's also the default) to the list of domains available in the domain dropdown in viewlogs, only for super admins. This allows the admin to see all log entries even if they're not associated to a domain.

This is useful not only to see logs of all domains, but also to be able to see log entries regarding modification of admin users, and deletion of domains, which were previously not possible.

I added translations to some languages, and left an english default for the rest.

Please let me know if any changes are required.


DavidGoodwin:
Thanks for taking the time to contribute! Merged.

PR #515, opened 2021-07-15
reuse $incpath variable instead of dirname

vakartel:


DavidGoodwin:
Are there circumstances where $incpath would be correct, and dirname(__FILE__) wouldn't be?

vakartel:
just to not run filesystem-aware function dirname() second time because we have its result in $incpath variable

DavidGoodwin:
Oh right. I'm not sure that works in this instance - the code looks like : 

image


$incpath is defined outside of the function 'postfixadmin_autoload' - so wouldn't be visible inside it by default (you'd need to 'global' it).

I think the code also uses a $incpath somewhere else (in public/users?) so I'm not sure doing this is a good idea.

vakartel:
> $incpath is defined outside of the function 'postfixadmin_autoload' - so wouldn't be visible inside it

exactly

PR #520, opened 2021-07-23
Update Password_Expiration.md

zakthan:
This line should be like this:

        echo -e "Dear User, \n Your password will expire on ${RESULT[1]}" | mail -s "Password $INTERVAL days before expiration notication" -r $REPLY_ADDRESS  ${RESULT[0]} 

instead of 

        echo -e "Dear User, \n Your password will expire on ${RESULT[1]}" | mail -s "Password 30 days before expiration notication" -r $REPLY_ADDRESS  ${RESULT[0]} 

otherwise the subject will always be 30 days no matter what the $INTERVAL is


DavidGoodwin:
Thanks for taking the time to contribute! LGTM.

PR #536, opened 2021-08-31
Minor update: add port to mysql/mariadb

Sethox:
This change is needed be able to use $CONF['database_host'] (custom port) for mysql/mariadb


DavidGoodwin:
what would happen if $CONF['database_port'] is undefined? 

Perhaps add something to default it to 3306 if not specified ?

Sethox:
I made a revision based to your comment @DavidGoodwin.

DavidGoodwin:
Thanks @Sethox !

PR #541, opened 2021-09-13
For Turkish Servers

Osman-Azgin:
This scpipt crashing on servers which using Turkish language.I added set_locale() to common.php and my problem is solved.


DavidGoodwin:
Im not sure why you need to submit all the file permission change(s) in the pull request?

Wouldn't 

`setlocale(LC_ALL, 'en_US.UTF-8', 'en_US', 'en', 'english');`

Normally be handled by your server having the right locale set? 

Is the problem that we're assuming the locale is set to english, when it's not?

DavidGoodwin:
what sort of 'crashing' were you seeing before? Any sort of error message?

DavidGoodwin (changes requested):
1. There appear to be unnecessary file permission changes (0644 -> 0755)
2. There are some random CSS changes
3. There are some random comment additions in some of the PHP file(s)

Osman-Azgin:
My error log is here:
2021/09/13 22:59:17 [error] 3505196#0: *28 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught Error: Class 'Smarty_Internal_Undefined' not found in /usr/share/nginx/Apps/postfixadmin/lib/smarty/libs/sysplugins/smarty_internal_extension_handler.php:168
Stack trace:
#0 /usr/share/nginx/Apps/postfixadmin/lib/smarty/libs/Smarty.class.php(1067): Smarty_Internal_Extension_Handler->__get('loadPlugin')
#1 /usr/share/nginx/Apps/postfixadmin/lib/smarty/libs/sysplugins/smarty_internal_template.php(648): Smarty->loadPlugin('Smarty_Internal...')
#2 /usr/share/nginx/Apps/postfixadmin/lib/smarty/libs/sysplugins/smarty_template_compiled.php(183): Smarty_Internal_Template->loadCompiler()
#3 /usr/share/nginx/Apps/postfixadmin/lib/smarty/libs/sysplugins/smarty_template_compiled.php(141): Smarty_Template_Compiled->compileTemplateSource(Object(Smarty_Internal_Template))
#4 /usr/share/nginx/Apps/postfixadmin/lib/smarty/libs/sysplugins/smarty_template_compiled.php(105): Smarty_Template_Compiled->process(Object(Smarty_Internal_Template))
#5 /usr/share/nginx/Apps/postfixadmin/lib/sm" while reading response header from upstream, client: 213.74.86.71, server: postfixadmin.mysite.com, request: "GET /login.php HTTP/1.1", upstream: "fastcgi://unix:/run/php-fpm/www.sock:", host: "postfixadmin.mysite.com", referrer: "https://postfixadmin.mysite.com/main.php"

This error is due to Smarty_Autoloader class.
Line 96 ("$_class = strtolower($class);") produces "Smary_Internal.." instead of "smarty_internal.." if the server language is Turkish, so the class cannot be loaded

You should include the solution for this in the documentation instead of merge.

Osman-Azgin:
The file permission change(s) are my fault. Sory.

DavidGoodwin:
TIL - https://github.com/smarty-php/smarty/issues/155

and indeed : 

https://www.php.net/strtolower does say : 

`Note that 'alphabetic' is determined by the current locale. This means that e.g. in the default "C" locale, characters such as umlaut-A (Ä) will not be converted.`

DavidGoodwin:
Changing the locale within PHP doesn't feel like a very good solution though - as presumably it should respect the server's settings.

Ideally Smarty would merge their stuff; at a push we could merge the PR into the project but it's not ideal.

Osman-Azgin:
I understand. I just wanted to share because it took a lot of time to solve.
Thanks.

DavidGoodwin:
looks like it's been fixed on the smarty side of things - https://github.com/smarty-php/smarty/pull/586 - which should be in PostfixAdmin 3.3.12 (soon to be released).

PR #566, opened 2021-11-29
Add support for implicit TLS, replace "smtp_sendmail_tls" with "smtp_type"

### LINKED ISSUE #563: Support for implicit TLS
Enabling `smtp_sendmail_tls` in the configuration switches from plain text to `STARTTLS`, which is commonly used over port 25 (SMTP) and/or 587 (Submission).

It would be ideal to support implicit TLS, commonly used over port 465 (SMTPS). For reference: https://datatracker.ietf.org/doc/html/rfc8314

From [Roundcube](https://github.com/roundcube/roundcubemail/blob/a980375271bdb957da9e2b4f503616727f67a042/config/config.inc.php.sample#L43-L58):

```php
// SMTP server host (for sending mails).
// Enter hostname with prefix ssl:// to use Implicit TLS, or use
// prefix tls:// to use STARTTLS.
// Supported replacement variables:
// %h - user's IMAP hostname
// %n - hostname ($_SERVER['SERVER_NAME'])
// %t - hostname without the first part
// %d - domain (http hostname $_SERVER['HTTP_HOST'] without the first part)
// %z - IMAP domain (IMAP hostname without the first part)
// For example %n = mail.domain.tld, %t = domain.tld
// To specify different SMTP servers for different IMAP hosts provide an array
// of IMAP host (no prefix or port) and SMTP server e.g. ['imap.example.com' => 'smtp.example.net']
$config['smtp_server'] = 'localhost';

// SMTP port. Use 25 for cleartext, 465 for Implicit TLS, or 587 for STARTTLS (default)
$config['smtp_port'] = 587;
```

However, I would use `starttls://` for `STARTTLS` and `tls://` for implicit TLS, to avoid ambiguity.

davidebeatrici:
For reference: https://datatracker.ietf.org/doc/html/rfc8314

Please note that this only applies to the "send email" feature.

In the future we should implement it for the "fetch email" one too.

---

Fixes #563.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -167,13 +167,17 @@ $CONF['admin_name'] = 'Postmaster';
 $CONF['smtp_server'] = 'localhost';
 $CONF['smtp_port'] = '25';
 
+// The communication layer used.
+//
+// 'plain'    Everything in plain text (standard port: 25).
+// 'tls'      TLS/SSL from the very beginning (standard port: 465).
+// 'starttls' "STARTTLS" in plain text and then TLS/SSL (standard port: 587).
+$CONF['smtp_type'] = 'plain';
+
 // SMTP Client
 // Hostname (FQDN) of the server hosting Postfix Admin
 // Used in the HELO when sending emails from Postfix Admin
 $CONF['smtp_client'] = '';
-
-// Set 'YES' to use TLS when sending emails.
-$CONF['smtp_sendmail_tls'] = 'NO';
 
 // Encrypt - how passwords are stored/hashed in the database.
 //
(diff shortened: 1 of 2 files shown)


DavidGoodwin:
Thanks again ! :-)

davidebeatrici:
You're welcome! :-)

I noticed two different methods are used to access configuration string values:

- `$CONF['key']`
- `Config::read_string('key')`

Which one is preferred?

DavidGoodwin:
Probably Config::read_string() ... as it's type hinted to tell e.g psalm that a string is being returned. I doubt the code base is consistent though.

PR #584, opened 2022-01-12
view log pagination

bnchdan:
Hello,

I have added pagination to the view log page.

The request is made with the go2page function.

image

### DIFF
--- a/public/viewlog.php
+++ b/public/viewlog.php
@@ -99,13 +99,13 @@ if ($error != 1) {
 
     $where = [];
     $params = [];
-    if($fDomain) {
+    if ($fDomain) {
         $where[] = 'domain = :domain' ;
         $params['domain'] = $fDomain;
     }
 
     $where_sql = '';
-    if(!empty($where)) {
+    if (!empty($where)) {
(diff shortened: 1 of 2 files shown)


public/viewlog.php
```diff
@@ -44,11 +44,40 @@
...
+				break;
+			}
+		}
+	}
+	
+	if ($flag_fDomain == 0 ){
+		die('Unknown domain');
+	}	
+
+	$page_number_aux=escape_string($_GET['page']);
```
DavidGoodwin, line 66: I'd not use 'escape_string' here - i think you'd be best using e.g.

$page_number_aux = (int) ($_GET['page'] ?? 0);

(in more english like ... that's : give me the 'page' url parameter, and if it doesn't exist default it to zero; and cast the entire thing to a number)

DavidGoodwin:
Thanks for taking the time to contribute :) 

Aside from your formatting not being quite right, I think it looks good.

public/viewlog.php
bnchdan: I modified the line

bnchdan:
Thanks!

PR #592, opened 2022-02-09
Update PFACrypt.php

bnchdan:
If you have hashes like $nr$salt$hash in the admin table. I think you will not be able to migrate to {hash}algorithm (for example to {SHA512-CRYPT.B64}...).

I have added the following lines of code that check if the hash is like $1$ (md5) or $6$. If yes, the algorithm will be changed to 'SYSTEM'.

### DIFF
--- a/model/PFACrypt.php
+++ b/model/PFACrypt.php
@@ -60,6 +60,17 @@ class PFACrypt
                 $algorithm = $method_in_hash;
             }
         }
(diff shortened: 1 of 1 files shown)


DavidGoodwin:
Thank you for taking the time to contribute!

bnchdan:
Thanks!

PR #594, opened 2022-02-13
Add Mailbox quota translations

olicooper:
The edit mailbox quota page was showing the wrong quota limits. Added new translation items to fix this.


DavidGoodwin:
Thank you for this!

PR #596, opened 2022-02-19
NOT && NOT && NOT = at lest 1

ecxod:
the logic is wrong, if you want to say at least of a list of negations you must put the condition with AND


DavidGoodwin:
Well spotted; thank you!

PR #597, opened 2022-02-20
I started to prepare setup.php for i18n

ecxod:
i just started to introduce the short gettext tag to prepare the code for i18n internationalization.


DavidGoodwin (commented):
I don't think we should be including composer.phar in git/vcs
There is an 'install.sh' file that attempts to download it if it's not present etc.

DavidGoodwin (commented):
You're using gettext's _(...) behaviour for translation of strings.
For whatever legacy reason(s), PostfixAdmin has been using it's own language file(s) (e.g. languages/fr.lang) which are translated through the Smarty {$PALANG.foo}  lookup stuff.

setup.php was never written to use smarty template(s) so has never supported translation.

While I don't want to dismiss your efforts - I don't think it's a good idea for PostfixAdmin to have two means of translation.

Perhaps change the calls to _(...) in setup.php to be instead e.g _t(....) or something that is a thin wrapper around the existing language lookup code (see e.g. Config::lang(...) )  - so at least it can fit in with the existing translation/language switching logic, rather than duplicating?

PR #599, opened 2022-03-07
view mail logs

bnchdan:
Hi,

We are currently using Postfix Admin for deploying e-mail services and we think it is a great tool, but one feature that Postfix Admin doesn't have is that the admins cannot have access to their mail logs.

I added a new page 'mail log' to the Postfix Admin where you can download mail logs for a specific domain.
image
image
image

I have created the following:
            - cron directory - where will be a script that separates logs by domain ( it should be added in the crontab (maybe 24 h) or executed periodically )
            - maillogs directory - where logs will be saved
            - public/maillog.php  and templates/maillog.tpl
 Also, I have added/modified :
            - $CONF['maillogging'] in the config.inc.pub 
            - url_maillog = maillog.php  in the configs/menu.conf
            -  I added a new tab in the menu.tpl 

I don't know if this is the best approach, but this will be a great feature. 
It can be disabled if you set $CONF['maillogging']  to 'NO'.


DavidGoodwin:
Thanks for taking the time to write and submit a PR. I'll try and look at this in the next few days.

(It's a shame Postfix can't log directly to a database, as that would make things a lot better ... my main concern is the use of grep for parsing mail.log etc)

cron/postfix_logs.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
```
DavidGoodwin, line 2: cd $(dirname $0) ?

cron/postfix_logs.sh
```diff
@@ -0,0 +1,39 @@
...
+	    do
+
+		#check if folder for domain doesn t exists
+		if [ ! -d ../maillog/$domain ]
+		then
+    			echo "Directory does not exists for "$domain
+			#create direcotry
+			mkdir ../maillog/$domain			
+		fi
+
```
DavidGoodwin, line 26: with your use of mail.log.1 you presumably need to run this script fairly soon after logrotate has run?

public/maillog.php
```diff
@@ -0,0 +1,149 @@
...
+   
+   	if (isset($_GET['get_log'])){
+		// remove path from filename
+		if (strpos($_GET['get_log'], '/') !== false) { 
+    			die("Unknown file");
+		}
+		
+	
+		$file = '../maillog/'.$fDomain.'/'.$_GET['get_log'];
+
```
DavidGoodwin, line 66: If you had .... 
if (!file_exists($file)) { die(.....) ; } 
here, you could un-indent the header/readfile block, which might make it easier to read.

cron/postfix_logs.sh
bnchdan: Yes. The script will be run fairly soon after logrotate.

The admins can not have access to logs from the current day.  For the current day, the script should be executed maybe every one hour, and mail.log.1 changed to mail.log.

bnchdan (commented):
Also, can we add 'btn-block' to the main.tpl?
**before:**
image

**after:**
image

bnchdan (commented):
If the  $CONF["logging"]  is changed to "NO" in the config.inc.php, the view log will not be present in the menu (menu.tpl), but will appear on the main page (main.tpl) .
**before:**
image

**after:** 
image

cron/postfix_logs.sh
DavidGoodwin: OK. Perhaps I should look at rsyslog and how it puts stuff into a database....

bnchdan: I tried to move logs in the database, but in some cases will be a large table. For example(in some cases),  2.000.000 logs/24h(for each domain), and I don't know what will be the best method to allow an admin to search for a specific log.

bnchdan: > OK. Perhaps I should look at rsyslog and how it puts stuff into a database....

I don t know how to do that

bnchdan (commented):
@DavidGoodwin  should be default set to 'NO'?

bnchdan:
Do you think it is not the best approach? Should I close the pull request?

DavidGoodwin:
> @DavidGoodwin should be default set to 'NO'?

Probably?

cron/postfix_logs.sh
DavidGoodwin: I'd assume an admin would want to search for log messages to/from a specific address?

DavidGoodwin:
> Do you think it is not the best approach? Should I close the pull request?

I'm hoping other users will voice their opinions :-)

bnchdan (commented):
also, I added some glyphicon icons in the main.tpl  and menu.tpl

cron/postfix_logs.sh
bnchdan: yep. e-mail address and  maybe ip, datetime

Aurel035:
It would be a nice feature for mail domain admins to have access to their mail logs through postfixadmin.

drdelaney:
I am the opposite. I prefer ssh+grep+less to look at logs.
I know logs can be large, and having something split them up to new files and download manually seems like a few extra unnecessary steps.
This being said, others may like this, and as long as it can be turned off, it will just be a feature I won't use.

andreidobre22:
This is a very useful feature for Postfixadmin. This should be included in the project. At this moment I gave logs using a ftp server. 
It should be accesed from postfixadmin, not using some 3rd parties tools.

cron/postfix_logs.sh
DavidGoodwin: With regards to the volume of logs -> for a small server (like mine) - I suspect a MySQL/PostgreSQL database would cope with the volume fine. For someone with more logs, I'd hope they'd use something better like ElasticSearch.

I've just had a look at the 'rsyslog-pgsql' package (Debian).

The main problem with rsyslog-XXsql will be that the log entries in the systemevents table is just one row per line of the log file, with no extra structure ... so you'd (probably?) end up wanting to parse / process these records in order to get them more structured (doing a "select * From systemevents where syslogtag like 'postfix%' and message like '%from=<foo@bar.com%';" to find incoming mail from foo@bar.com isn't going to work well.

It would work fine if you know what time frame you're looking for and you've got a fairly distinct pattern (e.g "what postfix messages were logged involving icloud.com between 6:30 and 7pm last night?") .... but once you identify a matching line/message, you'd want to parse that to retrieve the postfix identifier, and then re-query the database to fetch the matching lines - e.g.

```
Syslog=# \x
Expanded display is on.
Syslog=# select * From systemevents where syslogtag like 'postfix%' and message like '%from=<david@exam%';
-[ RECORD 1 ]------+-----------------------------------------------------------------------------
id                 | 28
customerid         | 
receivedat         | 2022-03-21 21:54:13
devicereportedtime | 2022-03-21 21:54:13
facility           | 2
priority           | 6
fromhost           | mail
message            |  6DF2A86E5: from=<david@example.co.uk>, size=939, nrcpt=1 (queue active)
ntseverity         | 
importance         | 
eventsource        | 
eventuser          | 
eventcategory      | 
eventid            | 
eventbinarydata    | 
maxavailable       | 
currusage          | 
minusage           | 
maxusage           | 
infounitid         | 1
syslogtag          | postfix/qmgr[754655]:
eventlogtype       | 
genericfilename    | 
systemid           | 
-[ RECORD 2 ]------+-----------------------------------------------------------------------------
id                 | 35
customerid         | 
receivedat         | 2022-03-21 21:54:14
devicereportedtime | 2022-03-21 21:54:14
facility           | 2
priority           | 6
fromhost           | mail
message            |  50A32870C: from=<david@example.co.uk>, size=1692, nrcpt=2 (queue active)
ntseverity         | 
importance         | 
eventsource        | 
eventuser          | 
eventcategory      | 
eventid            | 
eventbinarydata    | 
maxavailable       | 
currusage          | 
minusage           | 
maxusage           | 
infounitid         | 1
syslogtag          | postfix/qmgr[754655]:
eventlogtype       | 
genericfilename    | 
systemid           | 

```


and then e.g. 

`select * From systemevents where syslogtag like 'postfix%' and message like ' 50A32870C%';`

DavidGoodwin:
@bnchdan why close?

neuffer:
A good question.

cron/postfix_logs.sh
bnchdan: I think with ElasticSearch you will need more resources. For example, if you have 1-2.000.000 logs/24h and you have 40-50 domains and you allow the postmasters to search logs in the last 30 days.  I think you will need to have a lot of RAM .

In my opinion, if we will create a folder for each domain where we will put the logs will be more simple and you will not have to give more resources.  Also, it will be more simple to configure.

bnchdan: Maybe a log listener that copies in 'real time' logs from ../var/log/mail.log in ../postfixadmin/maillog/domain/ will be a more better?

bnchdan: Maybe a log listener that copies in 'real time' logs from ../var/log/mail.log in ../postfixadmin/maillog/domain/ will be more better?

bnchdan: or every 1 hour/ 30 min

DavidGoodwin: e.g. https://superuser.com/questions/181517/how-to-execute-a-command-whenever-a-file-changes ?

bnchdan: > e.g. https://superuser.com/questions/181517/how-to-execute-a-command-whenever-a-file-changes ?

yes

bnchdan: https://github.com/bnchdan/postfixadmin_logs/blob/main/main.sh maybe? it is not yet finished

bnchdan: something like https://github.com/bnchdan/postfixadmin_logs/blob/main/main.sh  ?

public/maillog.php
```diff
@@ -0,0 +1,150 @@
...
+*/
+
+$fDomain=$list_domains[0];
+
+if ($_SERVER['REQUEST_METHOD'] == "GET") {
+    if (isset($_GET['fDomain'])) {
+        $fDomain_aux = escape_string($_GET['fDomain']);
+        $flag_fDomain = 0;
+	//check if domain exists
+        if ((is_array($list_domains) and sizeof($list_domains) > 0)) {
```
cboltz, line 38: `$list_domains` will always be an array (unless `list_domains()` or `list_domains_for_admin()` fails - but you should check for such an error earlier.

Also, the `sizeof()` check looks a bit superfluous - with an empty array, the foreach loop will simply run 0 times.

Speaking about the foreach loop in the next lines - unless I missed something, this looks like a manual implementation of https://www.php.net/in_array

public/maillog.php
```diff
@@ -0,0 +1,150 @@
...
+		ob_clean();
+		flush();
+		readfile($file);
+    			
+
+
+   	}
+
+    }
+} elseif ($_SERVER['REQUEST_METHOD'] == "POST") {
```
cboltz, line 86: The code handling POST requests does some checks (duplicated from the GET handling, which is an issue on its own - please de-duplicate), but then - nothing except setting `$fDomain` and `$flag_fDomain`. Is this really intentional?

public/maillog.php
```diff
@@ -0,0 +1,150 @@
...
+    die('Unknown request method');
+}
+
+
+$path = '../maillog/'.$fDomain; 
+  
+//check if folder exists
+$log_list='';
+$log_size_list=array();
+if (file_exists($path)){
```
cboltz, line 120: Since you expect `$path` to be a directory, https://www.php.net/is_dir might be a better check

public/maillog.php
```diff
@@ -0,0 +1,150 @@
...
+*/
+
+require_once('common.php');
+
+authentication_require_role('admin');
+
+$CONF = Config::getInstance()->getAll();
+$smarty = PFASmarty::getInstance();
+
+$PALANG = $CONF['__LANG'];
```
cboltz, line 15: Do you really need `$CONF` and `$PALANG`? Smarty gets them on its own, and you don't seem to use them otherwise.

(Besides that, reading only the config option you need is preferred.)

templates/main.tpl
```diff
@@ -3,34 +3,44 @@
 
     <table class="table">
         <tr>
-            <td nowrap="nowrap"><a class="btn btn-primary" href="{#url_list_domain#}">{$PALANG.pAdminMenu_list_domain}</a></td>
-            <td>{$PALANG.pMain_overview}</td>
+            <td nowrap="nowrap"><a style="text-align:left; padding-left:15px" class="btn btn-primary btn-block" href="{#url_list_domain#}"><span class="glyphicon glyphicon-th-large"
+                                                         aria-hidden="true"></span> {$PALANG.pMenu_overview}</a></td>
+            <td style="padding-top: 15px;">{$PALANG.pMain_overview}</td>
```
cboltz, line 8: Using `btn-block` and adding icons is a good idea, but unrelated to this (quite big) PR. May I ask you to submit that as a separate, small MR?

templates/menu.tpl
```diff
@@ -28,42 +30,42 @@
                     {if $authentication_has_role.global_admin}
                         {strip}
                             <li class="dropdown">
-                                <a class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true"
-                                   aria-expanded="false" href="{#url_list_admin#}">{$PALANG.pAdminMenu_list_admin} <span
+                                <a class="btn navbar-btn dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true"
+                                   aria-expanded="false" href="{#url_list_admin#}"><span class="glyphicon glyphicon-list" aria-hidden="true"></span> {$PALANG.pAdminMenu_list_admin} <span
```
cboltz, line 34: These changes (adding icons etc.) also look like something for a separate PR. (Feel free to combine it with the main.tpl changes PR.)

DavidGoodwin: probably part of https://github.com/postfixadmin/postfixadmin/pull/604 ?

cboltz: Would have been a good fit, but since #604 is already merged (but didn't include main.tpl and menu.tpl), it's too late ;-)

cboltz: Just found out that these two files to their icons added in #603 :-)

I guess this MR needs a rebase on latest master ;-)

public/maillog.php
```diff
@@ -0,0 +1,121 @@
...
+* 	        - remove old logs
+*/
+
+require_once('common.php');
+
+authentication_require_role('admin');
+
+if (!is_dir(__DIR__ . '/../maillog')) {
+    die("../maillog does not exist");
+}
```
cboltz, line 14: The script should also die() if `$CONF['maillogging']` is not enabled.

BTW: config.inc.php should also set a default value for it (I'd tend to NO).

languages/en.lang
```diff
@@ -67,7 +67,7 @@ $PALANG['pMain_create_alias'] = 'Create a new email forward.';
 $PALANG['pMain_create_mailbox'] = 'Create a new mailbox.';
 $PALANG['pMain_sendmail'] = 'Send an email to one of your newly created mailboxes.';
 $PALANG['pMain_password'] = 'Change the password for your admin account.';
-$PALANG['pMain_viewlog'] = 'View the log files.';
+$PALANG['pMain_viewlog'] = 'View PostfixAdmin logs.';
```
cboltz, line 70: Note: You'll need to update all the `*.lang` files, either by updating the texts (if you speak 35 languages ;-) or at least by letting `language-update.sh` add a TODO note.

templates/menu.tpl
```diff
@@ -30,42 +28,67 @@
                     {if $authentication_has_role.global_admin}
                         {strip}
                             <li class="dropdown">
-                                <a class="btn navbar-btn dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true"
-                                   aria-expanded="false" href="{#url_list_admin#}"><span class="glyphicon glyphicon-list" aria-hidden="true"></span> {$PALANG.pAdminMenu_list_admin} <span
+                                <a class="btn navbar-btn dropdown-toggle" data-toggle="dropdown" role="button"
+                                   aria-haspopup="true"
+                                   aria-expanded="false" href="{#url_list_admin#}"><span
+                                            class="glyphicon glyphicon-list"
+                                            aria-hidden="true"></span> {$PALANG.pAdminMenu_list_admin} <span
```
cboltz, line 35: All these whitespace changes make this PR hard to read. Maybe you can split them out to a separate PR?

Legoman99573:
I love this implementation, but I would like to see the log files display in browser over downloading the files, so maybe adding a view button would help out and be a win/win as it would give admins an option to choose.

frzqwerty:
I love this implementation as well, but too sad, it will not work with OpenBSD.

OpenBSD's httpd (and even nginx from the packages) chrooted into /var/www. So, it could only access to /var/www/logs, instead of dovecot's/postfix's log path which is /var/log... :(

PR #600, opened 2022-03-08
update mysql_transport_maps.cf - translate 'virtual' to ':'

cantoute:
This makes sure our internal/virtual domains aren't caught in other transport map/nexthop

I needed this to safely implement transport rules like .fr => relay:[smtp.relay.com] without smashing our internal domains

So found out that transparent transport was ':' but took me a wile, so thought I could share.


DavidGoodwin:
This isn't functionality I use, so I'll have to hope it works; Thank you for contributing!

PR #606, opened 2022-03-21
add check box to domains

bnchdan:
Hi,

Can we add a check box to the domains list when you add a new admin? 

(Now if you want to add an admin to multiple domains you need to press ctrl )

before:
image

after: 
image


DavidGoodwin:
LGTM. Thank you.

PR #607, opened 2022-03-21
search domains

bnchdan:
Also, can we add a search bar?
I think is useful if you have multiple domains.

image
image


bnchdan:
thanks!

PR #609, opened 2022-03-21
Patch 2

bnchdan:
I hurried...
and I have forgotten that you can edit a user. and the check box will not appear checked.
image


bnchdan:
next time I will be more careful

PR #611, opened 2022-03-25
Update it.lang

freecris:
Translated some more lines


DavidGoodwin:
Thank you for taking the time to do this!

freecris:
Thank you to reply so fast!
Next week will finish that incomplete translation

Il Ven 25 Mar 2022, 17:03 David Goodwin ***@***.***> ha
scritto:

PR #639, opened 2022-07-15
Test against PHP 8.1 too

PowerKiKi:
This enable CI against PHP 8.1 too and fix a trivial deprecation message


DavidGoodwin:
Thank you!

PR #641, opened 2022-07-24
Feature save in transaction

DavidGoodwin:


model/PFAHandler.php
```diff
@@ -590,20 +590,29 @@ public function save(): bool
...
             }
+
+            // problems with this -> postSave, if it executes a db query, will not see the updated db values until this
+            // transaction has been committed.
+            $result = $this->postSave();
+
+            // commit db transaction
+            $pdo->commit();
+
         } catch (PDOException $e) {
```
DavidGoodwin, line 610: should we catch any Exception here instead?

PR #664, opened 2022-10-12
update languages/nl.lang

Jan-Kruis:
languages/nl.lang	 		translation from english for new function such as dkim and password
public/editactive.php 		edit wrong description text it said this module is used for deleting admin domains mailbos etc
						modify templates that it is used

public/list-virtual.php 	       modify templates that it is used
					       modify the remarks behind arguments and internal , the texteditor thought the text after this was still remark text untl is see */

public/vacation.php and
model/VacantionHandel.php 	adjust date string actionFrom and activeUntil these are written in the database as e.g.   2020-10-01 15:14:00 and 2020-10-30-15:14:00
							this should be 2020-10-01 00:00:00 and 2020-10-30 23:59:59 so that it contains all day, this happens if no holiday has been set for the account yet
							when adjusting the dates, the seconds of until are set to 00 instead of 59. When the holiday is canceled and then action is taken again, the from time
							to the time when the setting is made.

templates/editform.tpl remove blank space


DavidGoodwin:
why did you close the PR? 

For what it's worth, I would have preferred the language updates to be separate from the changes to the vacation date / time handling.

Jan-Kruis:
sorry i pressed the wrong button

PR #665, opened 2022-10-12
small correction

Jan-Kruis:
update for NL.lang

update ./model/VacationHandler.php and ./public/vacation.php to put correct date in ActiveFrom and ActiveUntil it wil now strore  YYYY-mm-dd 00:00:00 and YYYY-mm-dd 23:59:59 .

public/editactive.php      had the wrong description of the function

public/list-virtual.php     change template refence and adjusting remarks because texteditor sees rest of lines as remarks

templates/editform.tpl     remove extra space chars
templates/list-virtual_mailbox.tpl    remove extra indent


DavidGoodwin:
As I Said on #664 I would have preferred the vacation time change stuff to be on a different PR than the language update - as they're kind of two different things.... (plus I can't argue about the nl.lang ;-) )

Jan-Kruis:
Sorry,
your previous message came to me too late then I had already sent it again.

PR #670, opened 2022-10-28
Sendonly

Jan-Kruis:
send only

allows the administrator to make a mailbox sendonly or not, for example a noreply mailbox

the following file have been changed

DOCUMENTS/DOVECOT.txt
DOCUMENTS/POSTFIX_CONF.txt
DOCUMENTS/Postfix-Dovecot-Postgresql-Example.md

config.inc.php
configs/menu.conf

languages/en.lang
languages/nl.lang

model/MailboxHandler.php
public/upgrade.php
templates/list-virtual_mailbox.tpl

public/editsendonly.php

also the sql query for dovecot and if desired also postfix must be adjusted

for dovecot see DOCUMENTS/DOVECOT.txt
for postfix see DOCUMENTS/Postfix-Dovecot-Postgresql-Example.md

for a quick adjustment of the sendonly status a file has been added to <POSTFIXADMIN_ROOT_DIR>/public/

editsendonly.php this is derived from editactive.php

in config.inc.php or config.local.php you can indicate at $CONF[sendonly] whether you want to use the function sendonly or not

the variable sendonly is placed in the table 'mailbox' after the variable active

### DIFF
--- a/DOCUMENTS/DOVECOT.txt
+++ b/DOCUMENTS/DOVECOT.txt
@@ -3,6 +3,7 @@
 # Originally written by: Massimo <AndyCapp> Danieli
 # Revised by: Sampsa Hario <shario> for Dovecot v1.0
 # Revised by: David Goodwin <david@palepurple.co.uk> for Dovecot 2.1.x  (2014/01/02)
+# Small adjustment by: Jan Kruis < jan@crossreference.nl> in the sql question regarding sendonly (2022/10/28)
 #
 
 More complete Dovecot documentation:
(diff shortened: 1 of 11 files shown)


gnanet:
I was managing mailservers for more than a decade long, and i would like you to consider some of my cases, and thoughts:
1. We had besides sendonly, also receiveonly mailboxes (for suspended customers, also a receiveonly + disabled login)
2. a sendonly like noreply should not REJECT, in case someone is protecting itself with a quick test for return delivery, i would rather opt for DISCARD.
3. BUT, if you discard for sendonly, every mailer-daemon kind of problem has to detected during initial sending, because any later feedback from mailer-daemon would be discarded too
4. To avoid the REJECT/DISCARD problems, we wrote a rather spartan noreply-handler, which got all the mails from noreply@ account piped, and parsed the email, so in case it was a recognized mailer-daemon we could store it, or better, we could set a temp or permanent undeliverable status on recipients of the original email.

You see, the sample case of "sendonly like noreply@" is mostly solved by other ways, so i would suggest to find cases that better fit as a sendonly mailbox.

One example i could think is to enable different sender accounts for the same mailaddress, so one can also control which user is alowed to send (by permitting the sender account to use that email as sender address)

Anyway i like some of your solutions, for example the sql query with REJECT / OK based on sender only

PR #677, opened 2022-11-04
Forward admin users merge

Jan-Kruis:
.../public/users/edit-alias.php renamed to .../public/edit-forward.php and modified so admins can use it too.
As was done at .../public/user/vacation.php and .../public/vacation.php
 
.../public/users/edit-forward.php added.
This provides a reference to .../public/edit-forward.php.
.../public/edit-forward.php uses .../templates/edit-forward.tpl as template this was .../templates/edit-alias.tpl

This allows .../public/user/edit-alias.php and .../templates/edit-alias.tpl to be dropped

Made changes in .../templates/list-virtual_mailbos.tpl , .../templates/user_menu.tpl and .../config/menu.conf so that they now cause .../public/edit-forward .php is called for both users and admins.

TODO
        .../public/edit-forward.php to use templates/editform.tpl at David Goodwin's request to reduce the number of templates.

### DIFF
--- a/languages/nl.lang
+++ b/languages/nl.lang
@@ -288,7 +288,6 @@ $PALANG['description'] = 'Omschrijving';
 $PALANG['aliases'] = 'Aliassen';
 $PALANG['pAdminList_domain_quota'] = 'Domein quota (MB)';
 $PALANG['pAdminList_domain_backupmx'] = 'Back-up MX';
-$PALANG['pAdminList-tls_policy'] = 'TLS Policy';
 $PALANG['last_modified'] = 'Laatst bewerkt';
 
 $PALANG['pAdminCreate_domain_welcome'] = 'Voeg een nieuw domein toe';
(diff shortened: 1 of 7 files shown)


languages/nl.lang
```diff
@@ -319,7 +318,7 @@ $PALANG['pAdminEdit_domain_quota'] = 'Domein Quota';
 $PALANG['transport'] = 'Transport';
 $PALANG['backupmx'] = 'Backup-MX';
 $PALANG['pAdminEdit_domain_transport_text'] = 'Definieer transport';
-$PALANG['pAdminEdit_domain_backupmx'] = 'Definieer of deze Mailserver een Backup MX is voor het domain';
+$PALANG['pAdminEdit_domain_backupmx'] = 'Mailserver is Backup MX';
```
DavidGoodwin, line 321: are you sure you want to revert this NL to EN ?

templates/edit_forward.tpl
```diff
@@ -0,0 +1,50 @@
...
+                </div>
+            </div>
+        </div>
+        <div class="panel-footer">
+
+            <div class="btn-toolbar">
+                <div class="pull-right">
+                    <a href="main.php" class="mr btn btn-secondary">{$PALANG.exit}</a>
+
+                    <button class="ml btn btn-lg btn-primary" type="submit" name="submit" value="{$PALANG.save}">{$PALANG.save}</butt>
```
DavidGoodwin, line 44: </button> ??

DavidGoodwin (commented):
see comments...

languages/nl.lang
Jan-Kruis: yes, because the text is too long and does not cover the load well

Jan-Kruis:
david ,

there is more wrong, the correct template is not called
need to find out where and what went wrong

PR #749, opened 2023-07-28
toggle the style of the active buttons

reinks2112:
This is a small optical enhancement that toggles the style of the "active" buttons depending on the actual state of the corresponding entry.


DavidGoodwin:
Thank you for contributing :-)

PR #756, opened 2023-08-18
Fix default $smtp_client setting if we aren't connecting to the SMTP …

puck:
…server on localhost.

This fixes the issue noted on https://github.com/postfixadmin/postfixadmin/discussions/648


DavidGoodwin:
Thank you!

PR #764, opened 2023-09-21
View mail logs on postfixadmin

bnchdan:
View mail logs on postfixadmin using https://github.com/bnchdan/MailLog2MySQL
AUTH:
image
IMAP/POP3
image
SMTP:
image

### DIFF
--- a/templates/maillog.tpl
+++ b/templates/maillog.tpl
@@ -1,0 +1,326 @@
+
+<div>
+    <button class="btn" style="margin-bottom: 25px;"  onclick="changeTable('auth_logs')">   <span class="glyphicon glyphicon glyphicon-log-in">     </span> AUTH</button>
+    <button class="btn" style="margin-bottom: 25px;"  onclick="changeTable('dovecot_logs')"><span class="glyphicon glyphicon glyphicon-align-left"> </span> IMAP/POP3</button>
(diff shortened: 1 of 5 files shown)


templates/maillog.tpl
```diff
@@ -0,0 +1,326 @@
...
+            if (table == "dovecot_logs"){    
+                document.getElementById("content_logs").innerHTML+=`
+                    <tr>
+                        <td style="width:5%">`  +logs["data"][i]["month"]   +`</td>
+                        <td style="width:5%">`  +logs["data"][i]["day"]     +`</td>
+                        <td style="width:5%">`  +logs["data"][i]["hour"]    +`</td>
+                        <td style="width:20%">` +logs["data"][i]["domain"]  +`</td>
+                        <td style="width:20%">` +logs["data"][i]["email"]   +`</td>
+                        <td style="width:20%">` +logs["data"][i]["msgid"]   +`</td>
+                        <td style="width:20%">` +logs["data"][i]["log"]     +`</td>
```
DavidGoodwin, line 79: is this is going to be vulnerable to XSS ?

bnchdan: no. the data will be available just for the user.

neuffer:
Unfortunately this is MySQL only.
The author of the tool you are using is not interested in PostgreSQL.

PR #765, opened 2023-09-26
Javascript validation of username to prevent illegal characters

drunken-sod:
We've had problems with users entering a local_part containing characters later disliked by postfix or dovecot. This PR adds a javascript check on keystroke against a configurable regex and refuses to accept forbidden characters (now with helpful/annoying js-alert...).

There is no check at save as this is not a security-relevant issue, more of a service for the user.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -251,6 +251,13 @@ $CONF['password_validation'] = array(
     // 'length_check'          => function($password) { if (strlen(trim($password)) < 3) { return 'password_too_short'; } },
 );
 
+// Username legal characters
+// New/changed usernames will be checked against this regular expression with javascript
+// during entry, offending characters not displaying.
+// For example:
+// $CONF['username_legal_chars'] = '^[a-zA-Z0-9-_.]+$';
+$CONF['username_legal_chars'] = '';
+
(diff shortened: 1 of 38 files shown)


DavidGoodwin:
Thank you for contributing

cboltz:
I like the idea, but I'm not completely happy with the implementation - now we have two different ways of validation in the browser (this PR) and on the server side (`check_email()` in functions.inc.php) - including different regular expressions.

Can I convince you to adjust your JS code to use the regex from `check_email()`? ;-)

Also thanks for adding your new strings to the translation files. As a hint for future additions - the `language_update.sh` script makes this easier (add the new text to en.lang, and let the script update the other files).

For bonus points: We always mark not-yet-translated texts with `# XXX`. It would be nice if you do this in a follow-up commit/PR ;-)

drunken-sod:
Thanks for the merge @DavidGoodwin !
@cboltz , very valid point, I wanted it configurable as not all systems are standards compliant, not sure if the standard is compliant anymore ;-)
"mark not-yet-translated texts with # XXX" agh, yes, will do. Sorry.

PR #766, opened 2023-09-27
Add missing "# XXX" to failed translations

drunken-sod:
As @cboltz pointed out I had missed adding the "XXX" to failed translations defaulting to english in https://github.com/postfixadmin/postfixadmin/pull/765


DavidGoodwin:
Thank you!

PR #807, opened 2024-03-03
Update fr.lang

hguilbert:
Proposed translation update fr.lang


DavidGoodwin:
Thanks for taking the time to do this @hguilbert

PR #810, opened 2024-03-29
Update fr.lang

hguilbert:


DavidGoodwin:
Merci! (that's about the limit of my French, as my recent holiday in France confirms!)

cboltz:
Thanks for the translation updates!

To make the life of future translators (maybe even you) easier, can you please remove the `# XXX` comments on correctly translated texts? These comments are meant to mark texts that are not translated yet, or texts where the english text was changed. This also means the comments should be removed when translating a text.

PR #816, opened 2024-04-11
feat: support Dovecot DIGEST-MD5

bestlong:


.github/workflows/php.yml
```diff
@@ -44,12 +44,21 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Install Dovecot
+      run: |
+        sudo apt-get update -q
+        sudo apt-get install -yq dovecot-core
+        doveconf
+        cat /etc/passwd
```
DavidGoodwin, line 64: can you drop the cat / sudo lines please?

functions.inc.php
```diff
@@ -1076,11 +1077,14 @@ function _pacrypt_dovecot($pw, $pw_db = '')
...
 
-    # digest-md5 hashes include the username - until someone implements it, let's declare it as unsupported
+    $doveadm_options = '';
+
     if (strtolower($method) == 'digest-md5') {
-        throw new Exception("Sorry, \$CONF['encrypt'] = 'dovecot:digest-md5' is not supported by PostfixAdmin.");
+        if (empty($username)) {
+            throw new Exception("\$CONF['encrypt'] = 'dovecot:digest-md5' require username.");
+        }
+        $doveadm_options = ' -u ' . $username;
```
DavidGoodwin, line 1086: $username needs to pass through escapeshellarg?

functions.inc.php
```diff
@@ -1105,7 +1109,7 @@ function _pacrypt_dovecot($pw, $pw_db = '')
 
     $pipes = [];
 
-    $pipe = proc_open("$dovecotpw '-s' $method$dovepasstest", $spec, $pipes);
+    $pipe = proc_open("$dovecotpw '-s' $method$dovepasstest$doveadm_options", $spec, $pipes);
```
DavidGoodwin, line 1112: i think this needs rewriting to be a little clearer - e.g. 

.... {$method}{$dovepasstest}{$doveadm_options} ?

I wonder why -s has quote marks around it?

functions.inc.php
```diff
@@ -1323,9 +1327,10 @@ function _php_crypt_random_string($characters, $length)
  *
  * @param string $pw
  * @param string $pw_db optional encrypted password
+ * @param string $username optional
```
DavidGoodwin, line 1330: can you add something to the username comment to say it's currently only relevant for dovecot DIGEST-MD5 format

DavidGoodwin:
Thank you for taking the time to contribute to PostfixAdmin.

.github/workflows/php.yml
bestlong: It will clean up eventually, but sudo installation of dovecot is necessary.

functions.inc.php
bestlong: fix

DavidGoodwin (approved):
LGTM, thank you

PR #819, opened 2024-04-11
casting the date fields directly in the query

### LINKED ISSUE #818: vacation.pl - replace_string() is produced parsing errors
The replace_string introduced in commit be5dad4d0411c475c72375aed7e088bea57c2823 contains an hardcoded strptime() format string to parse activefrom and activeuntil dates. However, I'm using Postgresql, and my schema is laid out as:
```
 activefrom    | timestamp with time zone | default '2000-01-01 00:00:00+01'::timestamp with time zone
 activeuntil   | timestamp with time zone | default '2000-01-01 00:00:00+01'::timestamp with time zone
```

Which means that the results of the query will be:
```
postfix=# SELECT activefrom,activeuntil FROM vacation WHERE email='....';
       activefrom       |      activeuntil       
------------------------+------------------------
 2024-04-10 11:14:00+02 | 2024-04-13 13:14:00+02
(1 row)
```

Therefore the script will produce complaints:
```
 (garbage at end of string in strptime:  00:00:00+02 at /usr/lib64/perl5/Time/Piece.pm line 481, <STDI)
```

According to [perl_date_time.htm](https://www.tutorialspoint.com/perl/perl_date_time.htm) there's no way to catch the "+02" relative to the UTC offset.

Therefore I propose casting the field directly in the query, which solves the problem. I've validated the syntax with MySQL as well.

ntavares:
Fixes #818

### DIFF
--- a/VIRTUAL_VACATION/vacation.pl
+++ b/VIRTUAL_VACATION/vacation.pl
@@ -393,7 +393,7 @@ sub replace_string {
     my ($to) =@_;
     my $logger = get_logger();
 
-    my $query = qq{SELECT body,activefrom,activeuntil FROM vacation WHERE email=? };
+    my $query = qq{SELECT body,DATE(activefrom),DATE(activeuntil) FROM vacation WHERE email=? };
     my $stm = $dbh->prepare($query) or panic_prepare($query);
     $stm->execute($to) or panic_execute($query,"email='$to'");
     my @row = $stm->fetchrow_array;


DavidGoodwin:
on MariaDB/MySQL that's going to result in loss of the time - i.e.

```
MariaDB [postfixadmin]> SELECT body,DATE(activefrom),DATE(activeuntil) FROM vacation  \G
*************************** 1. row ***************************
             body: I will be away from <%From_Date> until <%Until_Date>.
For urgent matters you can contact <contact person>.
 DATE(activefrom): 2024-04-12
DATE(activeuntil): 2024-04-12
1 row in set (0.001 sec)
```

DavidGoodwin:
Thanks for taking time to contribute!

PR #823, opened 2024-04-21
dont assume 'quota' will be in the values, just like block before wasn't assuming

### LINKED ISSUE #822: Mailbox active toggle (editactive.php) crashes
Using version: [3.3.11] running on Postgres.

When clicking in the "active"  on/off link (to editactive.php), PHP will crash with:
 ```
Warning: Undefined array key "quota" in /var/www/html/model/MailboxHandler.php on line 231

Warning: Cannot modify header information - headers already sent by (output started at /var/www/html/model/MailboxHandler.php:231) in /var/www/html/public/editactive.php on line 61
```

Commit 864065cd37ef34b6dab915206eea4bd2ac4ebaed introduced this block without a check for isset($this->values['quota']) as it being checked the block just before.

My schema does have the field:
```
                                                 Table "public.mailbox"
     Column      |           Type           | Collation | Nullable |                      Default                       
-----------------+--------------------------+-----------+----------+----------------------------------------------------
[...]
 quota           | bigint                   |           | not null | 0
```

ntavares:
Fixes #822 . I just moved the offending block inside a shared "if isset()"

### DIFF
--- a/model/MailboxHandler.php
+++ b/model/MailboxHandler.php
@@ -239,17 +239,17 @@ class MailboxHandler extends PFAHandler
 
     protected function preSave(): bool
     {
-        if (isset($this->values['quota']) && $this->values['quota'] != -1 && is_numeric($this->values['quota'])) {
-            $multiplier = Config::read_string('quota_multiplier');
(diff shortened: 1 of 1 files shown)


DavidGoodwin:
Thank you for contributing!

PR #835, opened 2024-06-04
fix non escaped strings on languages/fr.lang

gnieark:


DavidGoodwin:
Merci! Thank you for spotting and fixing this.

PR #859, opened 2024-10-30
Fix deprecated config parameter error

kvassss:
The error appears when creating a mailbox. If the parameter is not specified in the config:
“PHP message: Config::read(): attempt to read undefined config option “create_mailbox_subdirs”, returning null”
If specified:
“PHP message: /opt/www/postfixadmin/model/MailboxHandler.php WARNING : PostfixAdmin no longer supports the imap folder population via config parameters, see https://github.com/postfixadmin/postfixadmin/issues/812”


ghost160111:
This is actually good!

PR #861, opened 2024-11-05
smtp_active should be a checkboox

gianlucagiacometti:
smtp_active field in the db is a boolean, so the form field should be a checkbox


DavidGoodwin:
Thank you for taking the time to contribute!

PR #864, opened 2024-11-08
Create new_mailbox_hook function

gianlucagiacometti:
This function may be useful to perform tasks on a newly created mailbox.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -372,6 +372,15 @@ $CONF['fetchmail_struct_hook']      = '';
 $CONF['dkim_struct_hook']           = '';
 $CONF['dkim_signing_struct_hook']   = '';
 
+/*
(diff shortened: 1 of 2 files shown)


DavidGoodwin:
Hi - Thanks for taking the time to contribute!

PostfixAdmin already has 'mailbox_postcreation_script' - which could technically do what you're after, but I can see you might want to run within the scope of PostfixAdmin instead.

I think it might be better if the config setting was renamed to **mailbox_postcreation_hook** rather than new_mailbox_hook.

gianlucagiacometti:
Hi,
I thank you for your constant effort and brilliant work.
I changed the name as you suggested, and I slightly modified the description in the config file.
I chose to create a function within the scope because it maybe useful, and sometimes necessary, to be able to use the record data.

PR #883, opened 2024-12-31
fixed non escaped strings - fr.lang

hguilbert:


hguilbert:
Continue what is done in the fr.lang file

e.g.
-$PALANG['pCreate_mailbox_email_desc'] = "Used in case of forgotten password";
+$PALANG['pCreate_mailbox_email_desc'] = 'Used in case of forgotten password';

But if this is useless, we might as well close the MR

DavidGoodwin:
OK, I wouldn't worry about the file being consistent in how the strings are quoted.  Either way is valid and both have their uses.

PR #921, opened 2025-05-17
avoid php deprecation message

buminks:
On Debian testing, postfixadmin-cli reports the following message:

Deprecated: exit(): Passing null to parameter #1 ($status) of type string|int is deprecated in /usr/share/postfixadmin/scripts/postfixadmin-cli.php on line 403

postfixadmin: 3.3.15
PHP: 8.4.6

This is maybe just a workaround, but it removed at least the deprecation warning.


DavidGoodwin:
Thanks @buminks - i think I'd prefer to use '??' than '||' - but the outcome is probably the same.
thanks for reporting this.

PR #922, opened 2025-05-26
Fixes 4.0 beta

DavidGoodwin:
random changes that may fix issues raised by @cboltz on IRC

 * table prefix not being respected for totp things
 * setup moaning about missing pgsql, but user doesn't care 
 * mysql tables need upgrading to innodb before some subsequent changes work

### DIFF
--- a/public/setup.php
+++ b/public/setup.php
@@ -622,22 +622,18 @@ function do_software_environment_check()
 
     if ($m_pdo_mysql) {
         $info[] = "Database - MySQL support available";
-    } else {
-        $info[] = "Database - MySQL (pdo_mysql) extension not found";
     }
 
-
-    // PostgreSQL functions
     if ($m_pdo_pgsql) {
         $info[] = "Database - PostgreSQL support available ";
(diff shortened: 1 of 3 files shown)


public/setup.php
```diff
@@ -620,24 +631,20 @@ function do_software_environment_check()
         $error[] = "There is no database (PDO) support in your PHP setup, you MUST install a suitable PHP PDO extension (e.g. pdo_pgsql, pdo_mysql or pdo_sqlite).";
     }
 
-    if ($m_pdo_mysql) {
-        $info[] = "Database - MySQL support available";
-    } else {
-        $info[] = "Database - MySQL (pdo_mysql) extension not found";
+    if (!$m_pdo_mysql && $CONF['database_type'] == 'mysql') {
+        $error[] = "Database support - MySQL compatible extension (pdo_pgsql) not found";
```
cboltz, line 635: That error should probably be `pdo_mysql` ;-) - and the database type for this case could also be `mysqli`

PR #944, opened 2025-10-07
Create mailbox_remover_pgsql.pl

VersusBG:
Mailbox remover for PostgreSQL database. Based on original mailbox remover.


VersusBG:
Yes, i can add some database handler selector and modify the original mailbox_remover.pl script to let the user to choose which database they use.

Yes instead of `if ($input =~ m/[Y]$/i) ` i could use `exit 0 if !($input =~ m/[Y]$/i);` but i just don't like to have exit code in the middle of the code that's why i prefer it to be inside the if statement.

PR #954, opened 2025-12-10
Re-implementation of the multiple server feature

Janfred:
This is a re-implementation of the multiple server feature ( #12 ) based on postfixadmin v4.

Useful for hot standby or backup systems with replicated databases.

The main problem is that, if you want to have a backup mx, you need to copy all mailbox setting changes you make on the first system to the second one.

With this new feature you can just replicate your database and use it for both, the primary and the secondary mailserver.

You can also do load balancing by sending one domain to the first and an other domain to the second mailserver as primary.

Most notable change from the last #12 version: Removed the post-create/post-delete scripts, as they may not be that useful after all.

TODO-List before merge:
* [ ] Add hint in `DOCUMENTS/BACKUP_MX.txt`
* [ ] Add a new document in `DOCUMENTS/` to describe this feature
* [ ] Modify example postfix scripts in `DOCUMENTS/POSTFIX_CONF.txt`

I will get to the TODO list if there is a response that this is a useful feature and will be merged.

### DIFF
--- a/public/upgrade.php
+++ b/public/upgrade.php
@@ -2412,3 +2412,39 @@ function upgrade_1851_mysql()
         )
     ");
 }
+
+function upgrade_1852_mysql_pgsql()
+{
+    # Add support for multiple servers
+    $server = table_by_key('server');
+    db_query_parsed("
+        CREATE TABLE {IF_NOT_EXISTS} $server (
+            `server` varchar(100) NOT NULL,
+            `description` varchar(255) NOT NULL,
+            `address` varchar(100) NOT NULL,
+            `created` {DATE},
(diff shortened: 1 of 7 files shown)


public/upgrade.php
```diff
@@ -2412,3 +2412,39 @@ function upgrade_1851_mysql()
         )
     ");
 }
+
+function upgrade_1852_mysql_pgsql()
+{
+    # Add support for multiple servers
+    $server = table_by_key('server');
```
DavidGoodwin, line 2419: why are there different functions for mysql + pgsql and sqlite? The SQL looks almost the same to me? (unless 'server' is a reserved word for sqlite?)

PR #977, opened 2026-01-25
Include domain in link to create new alias

csware:


templates/list.tpl
```diff
@@ -164,7 +164,11 @@
     <div class="btn-toolbar" role="toolbar">
         <div class="btn-group pull-right">
             {if $msg.can_create}
-                <a href="edit.php?table={$table|escape:"url"}" role="button" class="btn btn-default"><span
+                {assign var=tmpdomain value=""}
```
csware, line 167: Can there be cases there $fDomain is not set?

DavidGoodwin: yes, e.g. when viewing a list page where no domain has been deleted.

csware: Thanks! I haven't noticed that the very same tempalte is also used for the domain list.

DavidGoodwin: Yes, I was rather pleased I made the effort to test the code before blindly merging it !

Issue #924, opened 2025-06-09
[Suggestion / Feature Request] Reset TOTP


hguilbert:
Hello everyone,

In PostfixAdmin 4, how can one reset or disable the secret TOTP of a user/admin who has "lost" it ?

Is this option not available in postfixadmin-cli?

DavidGoodwin:
I can't see any way to do it in the UI - I suppose being an admin, it would make sense if you could edit/remove the totp setting on a mailbox if it's present?

hguilbert:
In fact, you can only modify the secret TOTP from the mailbox.

PR #988, opened 2026-02-24
WIP: TOTP reset for lost secrets

drunken-sod:
As pointed out by people who deal with users, there should be a way to clear TOTP secrets.
https://github.com/postfixadmin/postfixadmin/issues/924

- For some reason the templating layer insists on html encoding even if the struct entry is defined as html, therefore the decode in templates/editform.tpl.
- Maybe the actual reset action in edit.php should call the totp handler for extra permission checking.

### DIFF
--- a/public/edit.php
+++ b/public/edit.php
@@ -243,11 +243,14 @@ foreach ($form_fields as $key => $field) {
         } else {
             $fielderror[$key] = '';
         }
+        $sanitise = true;
+        if (isset($form_fields[$key]['type']) && $form_fields[$key]['type'] == 'html')
+            $sanitise = false;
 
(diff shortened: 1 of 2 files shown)


drunken-sod:
So in edit.php `$smarty->assign('struct', $form_fields);` the struct created in the admin/mailbox-handler is passed, including the field definitions i.e. "html" but sanitise() doesn't care. 
So I guess sanitise() should have something like
```
        if ((!is_array($data) && !is_string($data)) || (is_array($data) && $data['type'] == 'html' )) {
            return $data; // bool, int, null, object etc - can't sanitise.
        }
```

PR #1000, opened 2026-03-30
Add in-app update check and upgrade for global admins

knofte:
## Summary

Closes #998
Depends on #999

Adds a web-based update check and upgrade page accessible to global admins. The page queries the GitHub Releases API to show available updates with full changelogs from the running version, and allows one-click download and installation of updates.

## Features

- **Version check** — compares `$CONF['version']` against GitHub Releases API
- **Changelog** — shows release notes for all versions between current and latest
- **One-click upgrade** — downloads the self-contained release tarball and extracts it over the current installation
- **Safety checks** before applying:
  - Verifies PHP `PharData` extension is available
  - Checks the installation directory is writable
  - Never overwrites `config.local.php`
  - Clears OPcache after update so PHP serves the new files
  - Prompts admin to confirm before applying
  - Reminds to run `upgrade.php` for database migrations
- **Graceful fallback** — if no tarball asset exists (pre-#999 releases), links to the GitHub release page instead
- **Menu integration** — added to admin navbar, and footer "check update" link now points to this page for global admins

## How it works

1. Admin visits Update Check page (or clicks "check update" in footer)
2. Page fetches releases from `api.github.com/repos/postfixadmin/postfixadmin/releases`
3. Shows all newer versions with changelogs
4. If a self-contained `.tar.gz` asset is available (from #999), shows "Download & Apply" button
5. On click: downloads tarball to `sys_get_temp_dir()`, extracts via PharData, copies files over installation, resets OPcache
6. Redirects back with success message and link to `upgrade.php`

## Files changed

- `public/update-check.php` — new page with all update logic
- `templates/update-check.tpl` — Smarty template with Bootstrap 5 UI
- `configs/menu.conf` — added URL mapping
- `templates/menu.tpl` — added menu item for global admins
- `templates/index.tpl` — footer "check update" link points to new page for global admins

## Test plan

- [ ] Verify page loads and shows current version
- [ ] Verify GitHub API returns release list with changelogs
- [ ] Verify "up to date" message when running latest
- [ ] Verify writable/PharData checks display appropriate warnings
- [ ] Test upgrade flow with a self-contained release tarball (requires #999)
- [ ] Verify `config.local.php` is preserved during upgrade
- [ ] Verify OPcache is cleared after upgrade

### DIFF
--- a/public/update-check.php
+++ b/public/update-check.php
@@ -1,0 +1,417 @@
+<?php
+
+/**
+ * Postfix Admin
+ *
+ * LICENSE
+ * This source file is subject to the GPL license that is bundled with
+ * this package in the file LICENSE.TXT.
+ *
+ * Further details on the project are available at https://github.com/postfixadmin/postfixadmin
+ *
+ * @license GNU GPL v2 or later.
+ *
+ * File: update-check.php
+ * Checks for available updates via the GitHub Releases API and allows
+ * global admins to download and apply updates from the web UI.
+ *
+ * Template File: update-check.tpl
+ */
+
+require_once('common.php');
+
+authentication_require_role('global-admin');
+
+$smarty = PFASmarty::getInstance();
+
+$current_version = Config::read_string('version');
+
+/**
+ * Fetch content from a URL using cURL (preferred) or file_get_contents as fallback.
+ * Returns the response body as a string, or false on failure.
+ *
+ * @param string $url
+ * @param int $timeout
+ * @param array<string> $headers
+ * @return string|false
+ */
+function http_fetch(string $url, int $timeout = 10, array $headers = [])
+{
+    if (function_exists('curl_init')) {
+        $ch = curl_init($url);
+        if ($ch === false) {
+            return false;
+        }
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($ch, CURLOPT_FOLLOWLOCATION, true);
+        curl_setopt($ch, CURLOPT_TIMEOUT, $timeout);
+        curl_setopt($ch, CURLOPT_USERAGENT, 'PostfixAdmin-UpdateCheck');
+        if (!empty($headers)) {
+            curl_setopt($ch, CURLOPT_HTTPHEADER, $headers);
+        }
+        $response = curl_exec($ch);
+        $http_code = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+        curl_close($ch);
+        if (!is_string($response) || $http_code >= 400) {
+            return false;
+        }
+        return $response;
+    }
+
+    // Fallback to file_get_contents (requires allow_url_fopen)
+    $context = stream_context_create([
+        'http' => [
+            'method' => 'GET',
+            'header' => array_merge(['User-Agent: PostfixAdmin-UpdateCheck'], $headers),
+            'timeout' => $timeout,
+            'follow_location' => true,
+        ],
+    ]);
+
+    /** @psalm-suppress InvalidArgument */
+    return @file_get_contents($url, false, $context);
+}
+
+/**
+ * Fetch release information from the GitHub Releases API.
+ * Returns an array of releases or an empty array on failure.
+ */
+function fetch_github_releases(): array
+{
+    $response = http_fetch(
+        'https://api.github.com/repos/postfixadmin/postfixadmin/releases',
+        10,
+        ['Accept: application/vnd.github.v3+json']
+    );
+
+    if ($response === false) {
+        return [];
+    }
+
+    $releases = json_decode($response, true);
+    if (!is_array($releases)) {
+        return [];
+    }
+
+    return $releases;
+}
+
+/**
+ * Normalize a version tag to a comparable version string.
+ * Handles formats like "v4.0.1", "postfixadmin-3.3.16", "3.3.8".
+ */
+function normalize_version(string $tag): string
+{
+    // Strip common prefixes
+    $result = preg_replace('/^(postfixadmin-|postfxadmin-|v)/i', '', $tag);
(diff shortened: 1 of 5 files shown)


knofte:
## Local QA Results

| Check | Result |
|-------|--------|
| Lint PHP 8.4 | Pass |
| Lint PHP 8.5 | Pass |
| php-cs-fixer | 0 issues |
| Psalm (our code) | 0 issues (5 pre-existing in TotpPf.php) |
| PHPUnit 8.3 (sqlite + gd) | 74 tests, 976 assertions, all pass |
| PHPUnit 8.5 (sqlite + gd) | 74 tests, 976 assertions, all pass |

public/update-check.php
```diff
@@ -0,0 +1,417 @@
...
+ * Apply an update from a downloaded tarball.
+ * Extracts the archive over the current installation directory.
+ * Returns true on success, false on failure.
+ */
+function apply_update(string $tarball_path): bool
+{
+    $install_dir = dirname(__DIR__);
+    $tmp_dir = sys_get_temp_dir() . '/postfixadmin-update-' . uniqid();
+
+    if (!mkdir($tmp_dir, 0700, true)) {
```
cboltz, line 178: uniqid is predictable (time based) according https://www.php.net/uniqid which also means using it to create a tempdir could be a security risk.

A quick test shows that the `if !mkdir` can prevent symlink attacks (if the directory pre-exists as directory or symlink, mkdir will error out), so at least an attacker can "only" block the automated update.

Since I'm already commenting on this line - do you really think the `true` for recursively creating the directory is needed? IMHO it would be quite surprising if `sys_get_temp_dir()` points to a non-existing directory.

public/update-check.php
```diff
@@ -0,0 +1,417 @@
...
+            $install_dir = dirname(__DIR__);
+            if (!is_writable($install_dir)) {
+                flash_error('Installation directory is not writable by the web server. Cannot apply update.');
+            } else {
+                // Download to temp file (PharData needs .tar.gz extension to detect format)
+                $tmp_base = tempnam(sys_get_temp_dir(), 'postfixadmin-update-');
+                $tmp_file = $tmp_base ? $tmp_base . '.tar.gz' : false;
+                if ($tmp_base === false || $tmp_file === false) {
+                    flash_error('Failed to create temporary file.');
+                } elseif (!rename($tmp_base, $tmp_file)) {
```
cboltz, line 335: https://www.php.net/rename says that existing files will be overwritten. While it at least replaces an existing symlink (instead of giving a working symlink attack) (tested with PHP 8.4.19), this doesn't sound ideal. Maybe do the download into a temporary directory?

public/update-check.php
```diff
@@ -0,0 +1,417 @@
...
+                    flash_error('Failed to download update from: ' . htmlspecialchars($download_url));
+                    @unlink($tmp_file);
+                } elseif (apply_update($tmp_file)) {
+                    @unlink($tmp_file);
+                    // Clear OPcache so PHP serves the new files
+                    if (function_exists('opcache_reset')) {
+                        opcache_reset();
+                    }
+                    flash_info("Successfully updated to version " . htmlspecialchars($update_version) . ". Please run <a href='upgrade.php'>upgrade.php</a> to apply any database changes.");
+                    header('Location: update-check.php');
```
cboltz, line 348: Better send users directly to upgrade.php to enforce the database upgrade. Otherwise they might ignore the message ("hidden" in a success message) and in worst case hit random errors in various list or edit pages if an expected database change is missing.

Actually there's not really a need to send the users there - `require('upgrade.php');` should work.

(We only run check_db_version() on login.)

templates/update-check.tpl
```diff
@@ -0,0 +1,102 @@
+<div class="card mb-4">
+    <div class="card-header">
+        <h4><span class="bi bi-arrow-repeat" aria-hidden="true"></span> {$PALANG.pUpdate_check_title|default:'Update Check'}</h4>
+    </div>
+    <div class="card-body">
+        <p><strong>{$PALANG.pUpdate_current_version|default:'Current version'}:</strong> {$current_version}</p>
```
cboltz, line 6: Instead of using `|default:`, please use the `language-update.sh` script to add all needed texts to the language files.

public/update-check.php
```diff
@@ -0,0 +1,417 @@
...
+
+        if (empty($download_url) || empty($update_version)) {
+            flash_error('Missing update information.');
+        } elseif (!preg_match('#^https://github\.com/postfixadmin/postfixadmin/releases/download/#', $download_url)) {
+            flash_error('Invalid download URL — only official GitHub release assets are allowed.');
+        } else {
+            // Check the install directory is writable
+            $install_dir = dirname(__DIR__);
+            if (!is_writable($install_dir)) {
+                flash_error('Installation directory is not writable by the web server. Cannot apply update.');
```
cboltz, line 328: This check might be "good enough", but I can imagine several reasons why replacing some or all files might fail nevertheless:
- files not owned by the webserver user (+ maybe sticky bit set on a directory)
- subdirectories not owned by the webserver user
- AppArmor or SELinux preventing write access.

Doing a "perfect" check is hard (and probably impossible when it comes to checking AppArmor or SELinux permissions), so your "good enough" solution might indeed be good enough.

knofte:
Thanks for the feedback everyone. I've split this into two parts:

- **PR #1006** - version check + changelog display only (read-only, no self-update). Addresses cboltz's feedback on language strings. No config option needed as it degrades gracefully if HTTP is blocked.
- **This PR (#1000)** - the self-update feature, which can be revisited separately if there's appetite for it.

Suggest closing this one in favour of #1006 for now.

TrapoSAMA:
There is a better solution for this approach; I will create a PR with the proposal I have in mind.

PR #1007, opened 2026-04-06
Migrate TOTP exceptions to TotpexceptionHandler class

knofte:
As discussed with @cboltz on IRC - migrates TOTP exceptions from the custom totp-exceptions.php pages to a proper PFAHandler class, so they work with the standard list.php/edit.php infrastructure.

**What changed:**
- New `TotpexceptionHandler` class in `model/TotpexceptionHandler.php`
- Menu links updated to use `list.php?table=totpexception`
- User menu uses `../list.php?table=totpexception` for correct path resolution

**The Handler provides:**
- Standard CRUD via list.php/edit.php
- Role-based visibility (users see own + domain + global, admins see managed domains, global admins see all)
- IP validation in preSave
- Permission checks in preSave and delete
- Post-add/delete script execution (mailbox_post_totp_exception_add_script / _delete_script)

**Not removed yet:**
- The old totp-exceptions.php files (public/ and public/users/) are still present. Can be removed in a follow-up once this is confirmed working.

Two commits for easy review.

### DIFF
--- a/model/TotpexceptionHandler.php
+++ b/model/TotpexceptionHandler.php
@@ -9,6 +9,13 @@ class TotpexceptionHandler extends PFAHandler
     protected string $order_by = 'username, ip';
     protected ?string $user_field = 'username';
 
+    protected function no_domain_field()
+    {
+        // TOTP exceptions are not domain-scoped in the traditional sense.
+        // Visibility is handled in read_from_db_postprocess().
+        $this->allowed_domains = [];
+    }
+
     protected function initStruct()
     {
         $this->struct = array(
@@ -45,15 +52,15 @@ class TotpexceptionHandler extends PFAHandler
             'formtitle_edit' => 'pTotp_exceptions_welcome',
             'create_button' => 'pTotp_exceptions_add',
 
-            'required_role' => 'user',
+            'required_role' => 'admin',
             'listview' => 'list.php?table=totpexception',
             'early_init' => 0,
+            'user_hardcoded_field' => 'username',
         );
     }
 
     protected function validate_new_id()
     {
-        # auto_increment - any non-empty ID is an error
         if ($this->id != '') {
             $this->errormsg[$this->id_field] = 'auto_increment value, you must pass an empty string!';
             return false;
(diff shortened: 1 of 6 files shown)


model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+            'username'    => self::pacol(1,       1,      1,      'text', 'pTotp_exceptions_user',          ''),
+            'ip'          => self::pacol(1,       1,      1,      'text', 'pTotp_exceptions_address',       ''),
+            'description' => self::pacol(1,       1,      1,      'text', 'pTotp_exceptions_description',   ''),
+        );
+    }
+
+    protected function initMsg()
+    {
+        $this->msg['error_already_exists'] = 'pTotp_exception_result_error';
+        $this->msg['error_does_not_exist'] = 'pTotp_exception_result_error';
```
cboltz, line 48: I'd have to check where/in which cases these errors can appear, but I'm not sure if `Could not add exemption` fits both cases.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+    {
+        if (authentication_has_role('global-admin')) {
+            return parent::read_from_db($condition, $searchmode, $limit, $offset);
+        }
+
+        $table = table_by_key($this->db_table);
+        $username = $this->admin_username;
+
+        if (authentication_has_role('admin')) {
+            $domains = list_domains_for_admin($username);
```
cboltz: Using `$this->is_admin` and `$this->allowed_domains()` is easier ;-)

That said - this would be the first class to override read_from_db() (the existing classes use the function in PFAHandler).

Is there a good reason to override the function? (And would read_from_db_postprocess() be good enough to handle that?)

Note to myself: TODO: review the remaining part of this function if it needs to stay.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+        }
+        return $db_result;
+    }
+
+    protected function preSave(): bool
+    {
+        // Validate IP address
+        if (isset($this->values['ip']) && !filter_var($this->values['ip'], FILTER_VALIDATE_IP)) {
+            $this->errormsg['ip'] = Config::Lang('pException_ip_empty_error');
+            return false;
```
cboltz, line 147: Better use a `_validate_$fieldname` function to validate the content of a field.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        // Regular users can only add exceptions for their own username
+        if (!authentication_has_role('admin') && !authentication_has_role('global-admin')) {
```
cboltz: `if (!$this->is_admin) {` should be enough

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+            $this->values['username'] = $this->admin_username;
```
cboltz: admin_username is empty for non-admins, use `$this->username`

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        // Admins can only add exceptions for domains they manage
+        if (authentication_has_role('admin') && !authentication_has_role('global-admin')) {
```
cboltz: `if ($this->is_admin && !$this->is_superadmin) {`

Or check against $this->allowed_domains (which can also replace $domains in line +130)

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+            $exception_username = $this->values['username'] ?? '';
+            if ($exception_username !== $this->admin_username) {
+                if (strpos($exception_username, '@')) {
+                    list($_, $exception_domain) = explode('@', $exception_username);
+                } else {
+                    $exception_domain = $exception_username;
+                }
+                $domains = list_domains_for_admin($this->admin_username);
+                if (!in_array($exception_domain, $domains)) {
```
cboltz: just use $this->allowed_domains instead of $domains

Also, this could be done in `_validate_username()`

Looking at the whole function again, I wonder if it's logic is correct. You do this check only for admins, but not for superadmins, and also not for users. That doesn't sound too good...

BTW: Even for superadmins, checking the domain against $this->allowed_domains can't hurt - but you'll need to additionally allow an empty value for superadmins.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+        fclose($pipes[0]);
+        $output = stream_get_contents($pipes[1]);
+        fclose($pipes[1]);
+        $retval = proc_close($proc);
+
+        if (0 != $retval) {
+            error_log("Running $command yielded return value=$retval, output was: " . json_encode($output));
+            $this->errormsg[] = $warnmsg;
+            return false;
+        }
```
cboltz: Similar code to run a script probably exists multiple times in various handler classes (for example `MailboxHandler mailbox_post_script()`). It might make sense to have a common function (parameters: full command, warnmsg) that runs a given script.

With that, the calling function would "only" need to compose the full commandline.

BTW: Since PHP 7.4, the command can be an array - which means we don't have to do escape_shell_args() anymore.

(This is a candidate for a separate, small PR to adjust the existing script runs. Then rebase and adjust this PR when it's merged.)

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+    public function delete()
+    {
+        if (!$this->view()) {
+            $this->errormsg[] = Config::Lang($this->msg['error_does_not_exist']);
+            return false;
+        }
+
+        // Check permissions
+        $exception = $this->result;
+        $username = $this->admin_username;
```
cboltz: or `$this->username` for users

Looking at the following lines, better drop the variable and use $this->username / $this->admin_username directly

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+
+        if (!authentication_has_role('global-admin')) {
+            if (authentication_has_role('admin')) {
+                $domains = list_domains_for_admin($username);
```
cboltz: $this->is_admin / $this->is_superadmin / $this->allowed_domains

Also, to make the code easier to understand, I'd prefer

```
if ($this->is_superadmin) {
...
} elseif ($this->is_admin) {
...
} else {  # user
...
}
```

Note to myself: TODO: review remaining function after this is sorted out.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+            $proc = proc_open($command, $spec, $pipes);
+            if ($proc) {
+                fclose($pipes[0]);
+                $output = stream_get_contents($pipes[1]);
+                fclose($pipes[1]);
+                $retval = proc_close($proc);
+                if (0 != $retval) {
+                    error_log("Running $command yielded return value=$retval, output was: " . json_encode($output));
+                    $this->errormsg[] = Config::Lang('mailbox_post_totp_exception_delete_failed');
+                }
```
cboltz: another candidate for using a generic script runner function ;-)

model/TotpexceptionHandler.php
knofte: Reworked - now uses read_from_db_postprocess() instead of overriding read_from_db(), and uses $this->is_admin/is_superadmin/allowed_domains/username throughout. Also added _validate_ip() and _validate_username() methods, extracted run_post_script() helper, and normalizes empty username to NULL for global exceptions. Thanks for the detailed review.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
knofte: Addressed in rework.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
+<?php
+
+class TotpexceptionHandler extends PFAHandler
+{
+    protected string $db_table = 'totp_exception_address';
+    protected string $id_field = 'id';
+    protected string $label_field = 'ip';
+    protected ?string $domain_field = '';
+    protected string $order_by = 'username, ip';
```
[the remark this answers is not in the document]
knofte: Addressed in rework.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+        if ($limit > -1 && $offset > -1) {
+            $query .= " LIMIT $limit OFFSET $offset";
+        }
+
+        $db_result = array();
+        $result = db_query_all($query, $params);
+        foreach ($result as $row) {
+            $db_result[$row[$this->id_field]] = $row;
+        }
+        return $db_result;
```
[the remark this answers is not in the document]
knofte: Addressed in rework.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+     * - Users see exceptions for their username, domain, and global (NULL) exceptions
+     */
+    protected function read_from_db($condition, $searchmode = array(), $limit = -1, $offset = -1): array
+    {
+        if (authentication_has_role('global-admin')) {
+            return parent::read_from_db($condition, $searchmode, $limit, $offset);
+        }
+
+        $table = table_by_key($this->db_table);
+        $username = $this->admin_username;
```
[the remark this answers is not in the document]
knofte: Addressed in rework.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    public function webformConfig()
+    {
+        return array(
+            'formtitle_create' => 'pTotp_exceptions_welcome',
+            'formtitle_edit' => 'pTotp_exceptions_welcome',
+            'create_button' => 'pTotp_exceptions_add',
+
+            'required_role' => 'user',
```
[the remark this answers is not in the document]
knofte: Addressed in rework.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+            $domains = list_domains_for_admin($username);
+            $params = ['username' => $username];
+            $domain_conditions = [];
+            foreach ($domains as $i => $domain) {
+                $key = '_domain_' . $i;
+                $domain_conditions[] = "username = :$key";
+                $params[$key] = $domain;
+            }
+            $domain_sql = implode(' OR ', $domain_conditions);
+            $query = "SELECT * FROM $table WHERE username = :username OR $domain_sql ORDER BY $this->order_by";
```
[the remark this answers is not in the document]
knofte: Addressed in rework.

model/TotpexceptionHandler.php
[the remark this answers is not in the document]
knofte: Addressed in rework.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+            $warnmsg = Config::Lang('mailbox_post_totp_exception_add_failed');
+        } else {
+            return true;
+        }
+
+        if (empty($cmd)) {
+            return true;
+        }
+
+        $cmdarg1 = escapeshellarg($this->admin_username);
```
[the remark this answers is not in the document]
knofte: Addressed in rework.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+
+        // Regular users can only add exceptions for their own username
+        if (!authentication_has_role('admin') && !authentication_has_role('global-admin')) {
+            $this->values['username'] = $this->admin_username;
+        }
+
+        // Admins can only add exceptions for domains they manage
+        if (authentication_has_role('admin') && !authentication_has_role('global-admin')) {
+            $exception_username = $this->values['username'] ?? '';
+            if ($exception_username !== $this->admin_username) {
```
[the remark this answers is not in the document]
knofte: Addressed in rework.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     * Restrict visibility based on user role:
+     * - Global admins see all exceptions
+     * - Admins see exceptions for domains they manage
+     * - Users see exceptions for their username, domain, and global (NULL) exceptions
+     */
+    protected function read_from_db($condition, $searchmode = array(), $limit = -1, $offset = -1): array
+    {
```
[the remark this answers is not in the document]
knofte: Fair point on tests - will consider for a follow-up.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    /**
+     * Run a post-save/delete script via proc_open.
+     *
+     * @param string $command full shell command to execute
+     * @param string $warnmsg error message to display on failure
+     * @return bool true on success
+     */
+    private function run_post_script(string $command, string $warnmsg): bool
```
cboltz, line 283: This function can/should move to PFAHandler so that other Handler classes can also use it.

Also, please allow (or even enforce) that $command is an array so that we don't have to shell_escape all the parameters anymore.

(This can be done as a follow-up PR if you don't want to do it here.)

knofte: Done in PR #1009 - moved to PFAHandler::run_hook_script() with support for both string and array commands. Array form lets PHP handle escaping without escapeshellarg().

DavidGoodwin:
testing gives : 

```text
[08-Apr-2026 19:36:22 UTC] Config::read(): attempt to read undefined config option "totp_exception_address_struct_hook", returning null
[08-Apr-2026 19:36:27 UTC] Config::read(): attempt to read undefined config option "totp_exception_address_struct_hook", returning null
[08-Apr-2026 19:36:35 UTC] Config::read(): attempt to read undefined config option "totp_exception_address_struct_hook", returning null
[08-Apr-2026 19:36:35 UTC] Invalid query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'created' in 'INSERT INTO' caused by INSERT INTO `totp_exception_address` (username,ip,description,created,modified) VALUES (:username,:ip,:description,now(),now()) {"username":"fred","ip":"1.2.3.4","description":"testing"}
[08-Apr-2026 19:36:35 UTC] /....../model/PFAHandler.php - failed to save mailbox; message : SQLSTATE[42S22]: Column not found: 1054 Unknown column 'created' in 'INSERT INTO'
[08-Apr-2026 19:36:35 UTC] pTotp_exception_result_error used via read_f, but nothing replaced (value )

```

as the form said "or domain for admin" ... I also tried with a domain which triggered the same : 

```text
[08-Apr-2026 19:37:15 UTC] Config::read(): attempt to read undefined config option "totp_exception_address_struct_hook", returning null
[08-Apr-2026 19:37:15 UTC] Invalid query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'created' in 'INSERT INTO' caused by INSERT INTO `totp_exception_address` (username,ip,description,created,modified) VALUES (:username,:ip,:description,now(),now()) {"username":"example.com","ip":"1.2.3.4","description":"testing?"}
[08-Apr-2026 19:37:15 UTC] /...../model/PFAHandler.php - failed to save mailbox; message : SQLSTATE[42S22]: Column not found: 1054 Unknown column 'created' in 'INSERT INTO'
[08-Apr-2026 19:37:15 UTC] pTotp_exception_result_error used via read_f, but nothing replaced (value )
```

I've not yet looked into why the code thinks there must be a 'created' field.

public/upgrade.php
```diff
@@ -2183,3 +2183,40 @@ function upgrade_1851_mysql()
...
+    if (!_mysql_field_exists($table, 'modified')) {
+        db_query("ALTER TABLE $table ADD COLUMN modified datetime DEFAULT NULL");
+    }
+}
+
+function upgrade_1852_pgsql()
+{
+    $table = table_by_key('totp_exception_address');
+    if (!_pgsql_field_exists($table, 'created')) {
+        db_query("ALTER TABLE $table ADD COLUMN created timestamp DEFAULT NULL");
```
DavidGoodwin: other created fields are like : 

                created timestamp with time zone default now(),

or using the {DATETIME} place holder, which is equivalent to : 

                 timestamp NOT NULL default '2038-01-18

or sometimes {DATECURRENT} which is : 

                 timestamp NOT NULL default CURRENT_TIMESTAMP
                 timestamp with time zone default now() (pgsql).

So I think this should be changed to use the {DATETIME} or {DATECURRENT} placeholders instead - at least they'd be slightly more consistent ?

DavidGoodwin (dismissed):

public/upgrade.php
knofte: Fixed - simplified to use _db_add_field() with {DATECURRENT}, matching the existing convention.

model/TotpexceptionHandler.php
```diff
@@ -0,0 +1,314 @@
...
+        $this->allowed_domains = [];
+    }
+
+    /** @return void */
+    protected function initStruct()
+    {
+        $this->struct = array(
+            # field name              allow       display in...   type    $PALANG label                     $PALANG description  default / options / ...
+            #                         editing?    form    list
+            'id'          => self::pacol(0,       0,      1,      'num',  'pFetchmail_field_id',            '', '', array(), array('dont_write_to_db' => 1)),
```
DavidGoodwin: if this is merged ( https://github.com/postfixadmin/postfixadmin/pull/1008/changes/4cc2ee6a2de77ebb312c31008478fc680b991828 ) then this line needs to change

DavidGoodwin: (not sticking an array in that arg)

PR #1008, opened 2026-04-06
Add developer documentation for Handler classes

knofte:
As discussed with @cboltz on IRC - adds a developer guide for creating Handler classes.

Covers the full pattern: class naming, required properties, pacol/struct, field types, initMsg, webformConfig, validation methods, preSave/postSave, delete, read_from_db_postprocess, user mode, menu integration, and the request lifecycle.

Based on PFAHandler.php comments and lessons learned from building TotpexceptionHandler (#1007).

### DIFF
--- a/DOCUMENTS/HANDLER_CLASSES.md
+++ b/DOCUMENTS/HANDLER_CLASSES.md
@@ -101,7 +101,7 @@ protected function initStruct()
 | Type | Description |
 |------|-------------|
 | `text` | Text input |
-| `mail` | Email address (validated) |
+| `mail` | Email address (treated as text unless handler adds `_validate_*`) |
 | `pass` | Password (hashed via `pacrypt()` before storage) |
 | `b64p` | Base64-encoded password |
 | `num` | Numeric |
@@ -114,6 +114,10 @@ protected function initStruct()
 | `vnum` | Virtual numeric (not stored in DB) |
 | `vtxt` | Virtual text (not stored in DB) |
 | `quot` | Quota display |
+
+This list covers the most common types. Additional types may exist in
+`PFAHandler` and individual Handler classes — check the source for the
+full set.
 
 **Additional options via `$multiopt` array:**
 
@@ -173,8 +177,12 @@ public function webformConfig()
 }
 ```
 
-The `user_hardcoded_field` tells `list.php` / `edit.php` / `delete.php` which
-field to automatically set to the logged-in username when in user mode.
+The `user_hardcoded_field` must be set (non-empty) for `list.php` / `edit.php` /
+`delete.php` to allow user-mode access. It names the field that identifies the
+owning user, but note that the web entry points only check that this key is
+non-empty — they do not automatically set the field value. User scoping for
+SELECT queries is handled by `PFAHandler::$user_field`, and write enforcement
+must be done in the Handler itself (e.g. via `_validate_*` or `preSave()`).
 
 ### 7. Auto-increment IDs
 
@@ -297,7 +305,7 @@ Then reference them in templates using `{#url_foo#}`.
 
 ## Lifecycle
 
-1. `list.php`/`edit.php` instantiates the Handler: `new FooHandler($new, $username)`
+1. `list.php`/`edit.php` instantiates the Handler: `new FooHandler($new, $username, $is_admin)`
 2. Constructor calls `initStruct()`, `initMsg()`, sets up `$allowed_domains`
 3. For edit: `init($id)` loads the item, `set($values)` validates input, `save()` writes to DB
 4. `save()` calls `preSave()`, does the INSERT/UPDATE, then calls `postSave()`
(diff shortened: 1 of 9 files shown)


DOCUMENTS/HANDLER_CLASSES.md
```diff
@@ -0,0 +1,304 @@
...
+| `$default` | mixed | Default value |
+| `$options` | array | Options for `enum`/`list` types |
+| `$multiopt` | array or int | Named array for additional options, or `$not_in_db` as int |
+
+**Field types:**
+
+| Type | Description |
+|------|-------------|
+| `text` | Text input |
+| `mail` | Email address (validated) |
```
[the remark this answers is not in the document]
knofte: Fixed - clarified mail is treated as text unless handler adds validation.

DOCUMENTS/HANDLER_CLASSES.md
```diff
@@ -0,0 +1,304 @@
...
+| `num` | Numeric |
+| `bool` | Boolean (converted to database-specific boolean) |
+| `enum` | Dropdown from `$options` array |
+| `list` | Multi-select list |
+| `txta` | Textarea |
+| `txtl` | Text, displayed as list in list view |
+| `ts` | Timestamp (auto-formatted per locale) |
+| `vnum` | Virtual numeric (not stored in DB) |
+| `vtxt` | Virtual text (not stored in DB) |
+| `quot` | Quota display |
```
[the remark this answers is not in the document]
knofte: Added note that the list is non-exhaustive.

DOCUMENTS/HANDLER_CLASSES.md
```diff
@@ -0,0 +1,304 @@
...
+
+# user menu (note the ../ prefix for accessing public/ from public/users/)
+url_user_foo = ../list.php?table=foo
+```
+
+Then reference them in templates using `{#url_foo#}`.
+
+## Lifecycle
+
+1. `list.php`/`edit.php` instantiates the Handler: `new FooHandler($new, $username)`
```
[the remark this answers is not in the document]
knofte: Fixed - added $is_admin parameter.

DOCUMENTS/HANDLER_CLASSES.md
```diff
@@ -0,0 +1,304 @@
...
+        'listview' => 'list.php?table=foo',   // redirect target after edit
+        'early_init' => 0,                    // 1 = call init() before set()
+        'user_hardcoded_field' => 'username', // required if users can access
+    );
+}
+```
+
+The `user_hardcoded_field` tells `list.php` / `edit.php` / `delete.php` which
+field to automatically set to the logged-in username when in user mode.
+
```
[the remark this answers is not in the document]
knofte: Fixed - clarified that user_hardcoded_field is a gate for user-mode access, not an auto-setter. Write enforcement is the handler's responsibility.

model/PFAHandler.php
```diff
@@ -955,16 +960,15 @@ public static function pacol(int $allow_editing, int $display_in_form, int $disp
...
-            $not_in_db = 0; # keep default value
-            foreach ($multiopt as $key => $value) {
+        if (is_array($not_in_db)) { # remaining parameters provided in named array
+            trigger_error("PFAHandler::pacol() - passing mutiopt as an array is deprecated, please use the remaining parameters", E_USER_DEPRECATED);
+            foreach ($not_in_db as $key => $value) {
                 $$key = $value; # extract everything to the matching variable
             }
-        } else {
-            $not_in_db = $multiopt;
+            $not_in_db = 0; # keep default value
```
cboltz: The comment is wrong, the correct one would be `always use default`.

Now that the parameter is already called $not_in_db, this line hardcodes `$not_in_db = 0` which makes the parameter useless (and breaks some things). Please just delete this line (and the then-empty `else` block).

DavidGoodwin: if someone is still passing an array in as the $not_in_db parameter (was: `$multiopt`), then the $not_in_db parameter may not be set. The code was (incorrectly) trying to give $not_in_db a default value if an array had been passed in. 

Is this better? 

image

cboltz: I'm afraid as long as you rename the parameter to $not_in_db, it will always be set - and there's no way to tell the difference between an explizit 0 and a default 0. (And, worse - you'll also run into trouble if it's an array and contains a 'not_in_db' item - which results into overwriting the $not_in_db variable while using it in the loop.)

Therefore I'm afraid you'll have to revert your change so that the parameter name is not $not_in_db.

DavidGoodwin: i think i can move the check into the foreach loop or something like that; thanks for checking!

knofte:
Looks good - the named arguments approach is cleaner. Happy for this to be merged as-is.

model/PFAHandler.php
DavidGoodwin: see https://github.com/postfixadmin/postfixadmin/pull/1008/commits/1689fb062800967dd1efdb08962412f5edf37340

knofte:
@DavidGoodwin cboltz asked on IRC if you've tested pacol() with an array for $not_in_db, e.g.:

```php
pacol(..., not_in_db: ['not_in_db' => 1, 'linkto' => 'somewhere.php'])
```

He suspects the foreach loop that extracts array keys might break since the parameter was renamed. Could you add a test for this case?

PR #1014, opened 2026-04-14
Fix privilege escalation in TOTP exceptions and missing isset() in MFA

### LINKED ISSUE #1012: Global admin downgraded to regular admin in TotpPf.php
In `model/TotpPf.php`, the `addException()` and `deleteException()` methods check `authentication_has_role('admin')` before `authentication_has_role('global-admin')`. Since global admins have both roles in their session, they match the admin check first and get treated as regular admins (`$admin = 1` instead of `$admin = 2`).

This incorrectly restricts global admins to only managing TOTP exceptions for domains they explicitly manage, instead of having unrestricted access.

**Affected methods:**
- `addException()` (line ~315)
- `deleteException()` (line ~468)

**Fix:** Reorder checks to test `global-admin` first.

### LINKED ISSUE #1013: Missing isset() check on $_GET in users/login-mfa.php
In `public/users/login-mfa.php` line 38, `$_GET["abort"]` is accessed without an `isset()` check, causing a PHP Notice when the parameter is absent.

```php
if ($_GET["abort"] == "1" && authentication_mfa_incomplete()) {
```

The admin version at `public/login-mfa.php` line 48 correctly uses `isset()`:
```php
if (isset($_GET["abort"]) && $_GET["abort"] == "1" && authentication_mfa_incomplete()) {
```

knofte:
Two security fixes:

**1. Global admin privilege downgrade in TotpPf::addException() (fixes #1012)**

Global admins were incorrectly treated as regular admins because `authentication_has_role('admin')` was checked before `authentication_has_role('global-admin')`. Since global admins have both roles in their session, they matched the admin check first and were restricted to their own domains instead of having unrestricted access.

Fix: reorder to check `global-admin` first. Note: `deleteException()` already had the correct order.

**2. Missing isset() in users/login-mfa.php (fixes #1013)**

`$_GET["abort"]` accessed without isset() check, causing PHP Notice. The admin version at `public/login-mfa.php` already has the correct check.

Two separate commits.

### DIFF
--- a/model/TotpPf.php
+++ b/model/TotpPf.php
@@ -312,7 +312,10 @@ class TotpPf
             throw new \Exception(Config::Lang('pPassword_password_current_text_error'));
         }
 
-        if (authentication_has_role('admin')) {
+        if (authentication_has_role('global-admin')) {
+            $admin = 2;
+            // Global admins can do anything
+        } elseif (authentication_has_role('admin')) {
             $admin = 1;
             // Get domains the admin has access to
             $domains = list_domains_for_admin($username);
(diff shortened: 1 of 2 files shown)


DavidGoodwin:
thanks @knofte

knofte:
Happy to help :)

PR #1018, opened 2026-04-19
Fix PostgreSQL boolean binding in prepared statements

### LINKED ISSUE #1017: Error creating new mailbox
On branch master
Your branch is up to date with 'origin/master'.
all database updates are applied (version 1853) 

When trying to create a new mailbox I get following error:

`[Sun Apr 19 10:47:22.470256 2026] [proxy_fcgi:error] [pid 353207:tid 353347] [remote 2a02:3100:1bf1:8f00:35c7:75cc:dc05:3d83:40188] AH01071: Got error 'PHP message: Invalid query: SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type integer: "t"\nCONTEXT:  unnamed portal parameter $9 = '...' caused by INSERT INTO mailbox (username,local_part,domain,maildir,password,name,quota,active,smtp_active,phone,email_other,token,token_validity,created,modified,password_expiry) VALUES (:username,:local_part,:domain,:maildir,:password,:name,:quota,:active,:smtp_active,:phone,:email_other,:token,:token_validity,now(),now(),:password_expiry) {"username":"paperless_consume@example.com","local_part":"paperless_consume","domain":"example.com","maildir":"example.com\\/paperless_consume\\/","password":"$2y$10$6b75071cba322c6a7e925ulgZblGhonq\\/soLMH8jhZH3jHFsc4MYe","name":"Paperless Consume","quota":0,"active":"t","smtp_active":"t","phone":"","email_other":"","token":"","token_validity":"2026-04-19 08:47:22","password_expiry":"2026-04-19 08:47"}; PHP message: /var/www/html/vhosts/postfixadmin/model/PFAHandler.php - failed to save mailbox; message : SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type integer: "t"\nCONTEXT:  unnamed portal parameter $9 = '...'', referer: https://lyra.example.com/mailadmin/edit.php?table=mailbox`

knofte:
Fixes #1017

PostgreSQL rejects `'t'`/`'f'` strings when bound as PDO parameters to boolean columns, throwing `invalid input syntax for type integer`. These strings come from `db_get_boolean()` and work fine in raw SQL but fail in prepared statements.

**Fix:** In `db_query()`, detect `'t'`/`'f'` values on PostgreSQL and bind them with `PDO::PARAM_BOOL` instead of the default string type. All other values are bound normally.

**Why not change db_get_boolean()?** It's used in ~30 places including raw SQL concatenation in upgrade.php (DDL), CASE expressions in PFAHandler, and string comparisons in VacationHandler. Changing its return type would require touching all of those. Fixing the binding layer is safer and handles all callers at once.

### DIFF
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -1749,7 +1749,22 @@ function db_query(string $sql, array $values = array(), bool $ignore_errors = fa
     $stmt = null;
     try {
         $stmt = $link->prepare($sql);
-        $stmt->execute($values);
+        // PostgreSQL needs explicit boolean type binding for 't'/'f' values
+        // from db_get_boolean(), as PDO otherwise sends them as strings which
+        // fail with "invalid input syntax for type integer"
+        if (db_pgsql() && !empty($values)) {
+            foreach ($values as $key => $value) {
+                $param = is_int($key) ? $key + 1 : $key;
+                if ($value === 't' || $value === 'f') {
+                    $stmt->bindValue($param, $value === 't', PDO::PARAM_BOOL);
+                } else {
+                    $stmt->bindValue($param, $value);
+                }
+            }
+            $stmt->execute();
+        } else {
+            $stmt->execute($values);
+        }
     } catch (PDOException $e) {
         $error_text = "Invalid query: " . $e->getMessage() . " caused by " . $sql;
         error_log($error_text);
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -1606,6 +1606,30 @@ function db_get_boolean($bool)
 }
 
 /**
+ * Convert PostgreSQL boolean strings ('t'/'f') from db_get_boolean() to
+ * PHP native booleans for correct PDO parameter binding.
+ *
+ * PDO sends 't'/'f' as string parameters, which PostgreSQL rejects for
+ * boolean columns with "invalid input syntax for type integer".
+ *
+ * @param array $values parameter values for a prepared statement
+ * @return array values with 't'/'f' converted to true/false on PostgreSQL
+ */
+function _db_fix_pgsql_booleans(array $values): array
+{
+    if (db_pgsql()) {
+        foreach ($values as $key => $value) {
+            if ($value === 't') {
+                $values[$key] = true;
+            } elseif ($value === 'f') {
+                $values[$key] = false;
+            }
+        }
+    }
+    return $values;
+}
+
+/**
  * Returns a query that reports the used quota ("x / y")
  * @param string column containing used quota
  * @param string column containing allowed quota
@@ -1719,6 +1743,8 @@ function db_execute(string $sql, array $values = [], bool $throw_exceptions = fa
 {
     $link = db_connect();
 
+    $values = _db_fix_pgsql_booleans($values);
+
     try {
         $stmt = $link->prepare($sql);
         $stmt->execute($values);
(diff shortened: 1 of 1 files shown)


functions.inc.php
```diff
@@ -1749,7 +1749,22 @@ function db_query(string $sql, array $values = array(), bool $ignore_errors = fa
     $stmt = null;
     try {
         $stmt = $link->prepare($sql);
-        $stmt->execute($values);
+        // PostgreSQL needs explicit boolean type binding for 't'/'f' values
+        // from db_get_boolean(), as PDO otherwise sends them as strings which
+        // fail with "invalid input syntax for type integer"
+        if (db_pgsql() && !empty($values)) {
+            foreach ($values as $key => $value) {
```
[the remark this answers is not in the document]
knofte: Good catch - added the fix to db_execute() too via a shared _db_fix_pgsql_booleans() helper.

functions.inc.php
```diff
@@ -1749,7 +1749,22 @@ function db_query(string $sql, array $values = array(), bool $ignore_errors = fa
...
+                $param = is_int($key) ? $key + 1 : $key;
+                if ($value === 't' || $value === 'f') {
+                    $stmt->bindValue($param, $value === 't', PDO::PARAM_BOOL);
+                } else {
+                    $stmt->bindValue($param, $value);
+                }
+            }
+            $stmt->execute();
```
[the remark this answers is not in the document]
knofte: Fixed - now converts values in the array and passes to execute() normally, avoiding the bindValue type change issue.

functions.inc.php
```diff
@@ -1605,6 +1605,30 @@ function db_get_boolean($bool)
...
+ * boolean columns with "invalid input syntax for type integer".
+ *
+ * @param array $values parameter values for a prepared statement
+ * @return array values with 't'/'f' converted to true/false on PostgreSQL
+ */
+function _db_fix_pgsql_booleans(array $values): array
+{
+    if (db_pgsql()) {
+        foreach ($values as $key => $value) {
+            if ($value === 't') {
```
DavidGoodwin, line 1622: so if someone puts 't' in a field it'll get converted to a boolean?

knofte: In theory yes, but in practice the only source of bare 't'/'f' values in PostfixAdmin is db_get_boolean(). No text fields in the schema would contain a single 't' or 'f' as their full value.

The alternative is changing db_get_boolean() to return PHP native booleans, but that touches ~30 call sites including raw SQL in upgrade.php and CASE expressions. This felt like the safer fix.

Open to the other approach if you prefer.

knofte: To clarify - the check uses strict equality ($value === 't'), so it only matches a value that is exactly the single character 't' and nothing else. An email like t@example.com, a description, an IP address etc would never match. The only values that produce a bare 't' or 'f' in PostfixAdmin are from db_get_boolean().

DavidGoodwin: I think the code should change so the caller is passing a PHP boolean, PostgreSQL will be happy with that, and for other databases we either leave it as is (and it will get magically cast to an int) or we have some special case.

DavidGoodwin: see https://github.com/postfixadmin/postfixadmin/pull/1019/changes - is this simpler?

knofte:
Closing in favour of #1019 which takes the cleaner approach of fixing the source (passing native booleans) rather than patching at the binding layer.

PR #1030, opened 2026-05-08
Update appearance for consistency

Lombra:
- Make cards look closer to Bootstrap reference implementation by removing padding and using the proper header element.
- Vertically center align table row content.
- Add a theme adaptive background to the navigation bar, and reuse the design for non logged in pages.

bild
bild

### DIFF
--- a/templates/index.tpl
+++ b/templates/index.tpl
@@ -31,7 +31,7 @@
     {/if}
 {/if}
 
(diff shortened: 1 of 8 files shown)


DavidGoodwin:
Thanks for contributing / taking the time to create + submit the PR.

PR #1032, opened 2026-05-13
Fix pw recovery without totp

drunken-sod:
Check if user has TOTP active, if so display input field in pw-change form and evaluate it.

### DIFF
--- a/public/users/login-mfa.php
+++ b/public/users/login-mfa.php
@@ -36,7 +36,7 @@ if (authentication_has_role("user")) {
     exit(0);
 }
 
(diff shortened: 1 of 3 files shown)


drunken-sod:
The expanded condition in public/users/login-mfa.php‎ is not related to this bug, it just cluttered up the reporting and annoyed me with pointing out that the variable is undefined....

DavidGoodwin (approved):

knofte:
@drunken-sod Tack/Thanks, nicely sorted. :)

PR #1046, opened 2026-06-22
various patches from @trapoSama sent via email

DavidGoodwin:
* Keeps the assigned domain quota text and uses real mailbox usage for the quota bar color and fill.
 * Shows a red count next to a domain when one or more mailboxes exceed the configured quota alert threshold. Requires patch 1.
 * Adds optional configuration to hide mailbox quota and password expiry columns while retaining those values in the domain tooltip.
 * Corrects the quota-bar markup and aligns domain-list columns.
 * Replaces the quota bars in the domain overview and mailbox list with a unified bar component, preventing the fill, label, and border from separating. It hides Modified, Backup MX, and Transport progressively on reduced desktop widths, then uses horizontal scrolling as the final fallback. Requires patches 1 through 4.
 * Uses icon-only controls with translated tooltips in domain and mailbox lists: active state, SMTP active state, autoresponder, alias, edit, and delete. Active-state tooltips show the current value, for example `Activo: SI` or `Activo: NO`. Requires patch 5.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -804,6 +804,12 @@ $CONF['version'] = '4.0.1';
 // in your Dovecot SQL configuration.
 $CONF['smtp_active_flag'] = 'NO';
 
+// Domain list display.
+// Set these options to YES to hide optional columns in the domain list.
+// Hidden values remain available in the domain tooltip.
+$CONF['domain_list_hide_maxquota'] = 'NO';
+$CONF['domain_list_hide_password_expiry'] = 'NO';
+
 // If you want to keep most settings at default values and/or want to ensure
 // that future updates work without problems, you can use a separate config 
 // file (config.local.php) instead of editing this file and override some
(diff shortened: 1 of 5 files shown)


DavidGoodwin:
when `$CONF['domain_list_hide_password_expiry'] = 'YES' `

I see this error : 

broken domain listing

If you don't have xdebug installed, you'll probably just see an undefined variable somewhere in your PHP error log (at a guess).


I think the error is from this line - https://github.com/postfixadmin/postfixadmin/pull/1046/changes#diff-bd011298ea6d073bd67ceaa1cab366106b731a52815b944aefe4c0661c4142eeR67

TrapoSAMA:
I reproduced the issue from the reported configuration path. The tooltip code can reference optional hidden values that are not present in the row when domain_list_hide_password_expiry = YES.

I opened a small hotfix PR against this branch: #1048

It only adds isset() guards for maxquota and password_expiry before adding those values to the domain tooltip.

PR #1051, opened 2026-06-28
Polish Spanish labels and domain quota text

TrapoSAMA:
This PR contains the remaining local patches 0012-0014.

Changes:

 * Polish Spanish translations for mail, DKIM, logs, quota and Backup MX labels.
 * Use a shorter domain-list quota label while keeping the full domain quota label in edit forms.
 * Keep hidden domain-list values available in the domain tooltip without reusing form labels.
 * Add explicit labels for max mailbox quota and unmanaged quota state.
  
Validation:
 * Applied cleanly on current origin/master.
 * php -l passed for languages/en.lang, languages/es.lang and model/DomainHandler.php.
 * git diff --check


knofte:
Should likely be "Not Managed" instead of "not managed" in the languages though. Keeping the same case type.

TrapoSAMA:
Updated, thanks. I changed the unmanaged quota label to title case in both languages: English: Not Managed. Spanish: Sin Cuota Gestionada.

PR #1055, opened 2026-07-01
Handle expired sessions for invalid tokens

TrapoSAMA:
## Summary

This backports the invalid-token/session-expired UX improvement to the 4.0 branch.

When a session expires and an old CSRF token is submitted, PostfixAdmin currently shows a raw invalid token error. This change renders a translated session-expired page instead, returns HTTP 419, provides a login button, and redirects to login after 5 seconds.

## Validation

- php -l public/common.php
- php -l languages/en.lang
- php -l languages/es.lang
- git diff --check

### DIFF
--- a/languages/en.lang
+++ b/languages/en.lang
@@ -518,6 +518,10 @@ $PALANG['pLegal_char_warning']              = 'Illegal character';
 $PALANG['TOTP_already_configured']          = 'TOTP is already configured for this account, do you want to change your secret?';
 $PALANG['pApp_passwords_add']               = 'Add application-specific password';
 
+$PALANG['session_expired_title'] = 'Session expired';
+$PALANG['session_expired_message'] = 'Your session has expired. Please log in again.';
+$PALANG['session_expired_redirect'] = 'You will be redirected to the login page in 5 seconds.';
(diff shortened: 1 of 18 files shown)


DavidGoodwin:
(I've not tested this... but... ) 

I think you'll see the redirect-to-the-login-page behaviour if you open multiple tabs within postfixadmin and have clicked around in both of them, because we only store a single CSRF token, and it gets reset on each page load.

Given doing a GET on login.php results in the session being destroyed, this is probably a worse situation than before?

TrapoSAMA:
Thanks, good point. I updated the PR to avoid the automatic redirect to login.php.

The invalid-token page now returns HTTP 419 and shows a manual Continue button pointing to main.php. That keeps the current session intact for the multiple-tabs case, while an actually expired session will still follow the normal auth flow from main.php.

TrapoSAMA:
I found the same warning class while testing the invalid-token flow locally. On the 4.x branch the token checks use PFA_token directly, so missing session token values could emit PHP warnings before the friendly 419 response. I updated the token comparisons to use an empty fallback when the session value is missing. The latest GitHubBuild run is passing.

Issue #1057, opened 2026-07-01
Bulk-delete inactive forwards from the alias list


knofte:
### Enhancement

Follow-up to #1038 (a filter to hide/show inactive forwards). Once an admin can narrow the alias list to **inactive** forwards, the natural next step is clearing them out — but today deletion is strictly one row at a time, which is painful when a domain has accumulated many retired forwards over the years.

### Proposal

Add a **"Delete all inactive forwards"** action, shown only in the alias list's *Inactive* view. It deletes every inactive (`active = 0`) alias of the current domain in one confirmed POST.

This is deliberately **filter-driven** rather than a multi-select / "tick each row" UI:

- No new checkbox column, no select-all JavaScript, no changes to the shared `list.tpl`.
- The set of rows to delete is unambiguous (it's exactly what the Inactive filter already shows).
- Each alias is still deleted through `AliasHandler`, so the existing per-alias guards — domain ownership, protected/default aliases (`postmaster@` etc.), and mailbox aliases — are all honoured; those are skipped and reported, not force-deleted.

Scope is limited to aliases/forwards (the entity #1038 concerns); mailboxes are untouched.

### Notes

Active forwards are never affected. The action is POST-only and CSRF-protected, with a confirmation dialog. A PR implementing this (stacked on the #1038 filter) follows.

DavidGoodwin:
done

Issue #1038, opened 2026-05-29
[Enhancement] Selector to hide all !active aliases from lists.


jeffsf:
We've got quite a bit of back history on our domains to the point where previously active aliases outnumber active ones by a long shot. We'd like to keep a record of the "retired" accounts in the same database as the current ones, for ease of searching. However, this clutters the display for most tasks. 

A checkbox for "Hide inactive aliases" (or similar if appropriate for other entities) would make this a lot more manageable.

knofte:
Right, this is a good idea. And also perhaps a "mark all and delete" for a list of inactive ones.

DavidGoodwin:
done

PR #1058, opened 2026-07-01
Bulk-delete inactive forwards from the alias list (#1057)

knofte:
## What

Adds a **"Delete all inactive forwards"** button to the alias list's *Inactive* view, deleting every inactive (`active = 0`) forward of the current domain in one confirmed action.

Implements #1057. Builds on #1038 / #1056 (the inactive filter).

> **Depends on #1056** (the inactive filter). This branch is stacked on top of it, so the diff currently includes #1056's commit as well — the delete-specific change is the second commit (`Add filter-driven bulk delete of inactive forwards`). Easiest path: merge #1056 first and I'll rebase this onto master so it shows only its own commit. The button relies on the filter's `$alias_active` state, so it only does anything once #1056 is in.

## How

- New `public/delete-inactive.php`: POST-only, `CsrfToken::assertValid`, `authentication_require_role('admin')`. Validates the posted domain against `list_domains_for_admin()`.
- Enumerates the inactive aliases via `AliasHandler::getList(['domain' => …, 'active' => 0])` — already restricted to the admin's allowed domains — then deletes each through its own `AliasHandler`, so per-alias checks (ownership, protected/default aliases, mailbox aliases) are honoured. Skipped ones are counted and reported, never force-deleted.
- Button lives in `list-virtual.tpl`, gated to the alias `Inactive` view with a `confirm()` dialog. The shared `list.tpl` is untouched — no multi-select column, no select-all JS.
- 4 new `en.lang` strings (button, confirm, success/skip messages); other languages fall back to English.

Active forwards are never affected.

## Testing

- New unit test `AliasHandlerTest::testDeletingInactiveAliasesLeavesActiveOnes` — deletes the inactive set exactly as the endpoint does and asserts only inactive aliases are removed while active ones (and the active default aliases) remain. Full `AliasHandlerTest` passes (8 tests, 582 assertions) on the SQLite test DB.
- `php-cs-fixer` (dry-run) and `parallel-lint` clean on the new/changed files.
- Manual QA of the button + confirm + protected-alias skip path recommended on a populated DB.


knofte:
Addressed the Copilot findings on this PR: the hidden domain value is now `escape:"quotes"`, the `confirm()` string is `escape:"javascript"` (so a translated quote/newline can't break it), the four `delete_inactive_*` strings are added to all language files, and the test `@see` now points to #1057.

Note: the filter-link findings Copilot raised here (the `alias_active=...` links) belong to the stacked #1056 commit and are fixed there (validated `$fDomain|escape:"url"`). Once #1056 merges I'll rebase this branch onto master so the diff shows only the delete-specific changes.

DavidGoodwin:
@knofte I think there's a convention that untranslated strings have an `# XXX` comment after them (which is meant to make it easier for translators to see what strings need translating); 

It'd be easier if you just ran `language-update.sh --patch` after you modify the en.lang file.

knofte:
Done — the new untranslated strings in the non-English files now carry the `# XXX` marker (en.lang unchanged), same output as `language-update.sh --patch`. Will use the script after en.lang edits from now on.

DavidGoodwin:
@copilot resolve the merge conflicts in this pull request

DavidGoodwin:
merged in 94d8ffcf78a216c56643258b02816636bff62ea4

PR #1061, opened 2026-07-01
Let admins store a default for the alias "To" (goto) field

mkocustos:
## Summary

When creating an alias, the **To** (`goto`) field always starts empty. Admins who
routinely forward aliases to the same target — most often their own mailbox — end
up typing the same address over and over.

This PR adds an opt-in checkbox directly below the **To** field on the add-alias
form. Tick it and the destination just entered is stored as the default for future
aliases. The default behaviour is unchanged: as long as nothing is stored, the field
stays empty and the checkbox does nothing.

**This is a rework of the original approach**, following @DavidGoodwin's suggestion
of an inline control instead of a settings screen. The separate per-admin Settings
page is gone, along with its menu and dashboard entries, its `login`/`custom` prefill
modes and the mailbox-existence check that only existed to justify a dedicated screen.

## What it looks like

Nothing stored yet:

```
☐ Save this destination as the default "To" address for future aliases
```

Ticked (the label is rewritten on click to describe the state that was chosen):

```
☑ This destination will be saved as the default "To" address for future aliases
```

With a default already stored:

```
☐ Save this destination as the default "To" address for future aliases
☑ This destination will replace the saved default "To" address
   Currently saved: someone@example.org  ✕
```

The ✕ deletes the stored default straight away, without having to create an alias.

## Design notes

**Why persist instead of `$_SESSION`** — a session-scoped value is per browser and
gone on session expiry, so it does not survive moving between devices. Adding aliases
from both a laptop and a phone is the main use case, and @dsteinborn described the
same need on this PR.

**On "remember" being ambiguous** (@knofte's point — "do I have to tick this for the
alias itself to be saved?") — the label spells out the effect rather than naming it,
and changes on click to state what was chosen. The box is never pre-ticked, so
creating a one-off alias to a different address cannot silently overwrite a stored
default; replacing one is always a deliberate click.

**Why not `alias_struct_hook`** (@knofte) — that is configured in `config.local.php`
and therefore needs shell access to the server. This is aimed at domain admins who
only have the web UI and want to set their own prefill without involving a sysadmin:
self-service per-admin preference vs. global sysadmin-defined rule.

**On textareas and browser autocomplete** (@knofte) — the prefill does not rely on
browser value history at all. `PFAHandler` populates `$this->values[$key]` from
`$row['default']` in `$struct` before the form is rendered, which is the same generic
mechanism `editform.tpl` already uses for every field type, `goto` included. Since the
default is set in PHP before Smarty renders, it behaves the same whether the field is
an `<input>` or a `<textarea>`.

## How it works

- **Storage** — per-admin key/value rows in a new table `admin_preferences`
  (`username`, `pref_key`, `pref_value`), created by migration `upgrade_1855`. The
  DDL stays in the portable subset: `db_query_parsed()` substitutes the
  `{IF_NOT_EXISTS}` / `{COLLATE}` tokens per backend but does not rewrite identifier
  quoting or strip MySQL-only clauses, so backticks (rejected by PostgreSQL) and an
  inline `COMMENT='...'` (MySQL-only) are omitted; none of the identifiers are
  reserved words in MySQL, PostgreSQL or SQLite. `pref_value` is `varchar(2048)` so a
  full comma-separated `goto` list fits; it is not part of an index.
  `set_admin_pref()` uses delete+insert rather than an upsert, since upsert syntax
  differs across the three backends.

- **Prefill** — reuses the existing field-default mechanism:
  `AliasHandler::webformConfig()` sets `struct['goto']['default']` for new aliases, so
  `edit.php` renders the prefilled value with no change to the form controller. (For
  admins the login is read from `admin_username`, since `username` is only populated
  for non-admin mailbox users.)

- **The checkbox** — a `not_in_db` `bool` field displayed only when creating an alias,
  so aliases created by `MailboxHandler` leave the stored value alone. `editform.tpl`
  renders it through the per-table special-case hooks the template already provides;
  the changing label is ~14 lines of JS in the existing script block, and the feature
  works without it (the label simply stays on the unticked wording).

- **Deleting** — has to work without creating an alias, so it gets a small POST-only
  endpoint `public/alias-goto-default.php`, CSRF-checked in the same way as
  `delete.php`. The ✕ sits inside the alias form and submits a separate form placed
  after it via its `form` attribute, since forms cannot be nested.

- **i18n** — new strings in `de.lang` and `en.lang`; other locales fall back to the
  English keys.

## Testing

Verified against a copy of a production database (MariaDB), plus the migration on
SQLite:

- Nothing stored → no ✕ and no delete form in the markup, field empty.
- Tick + create → stored; next add-alias form is prefilled, the label switches to the
  "will replace" wording and the ✕ appears.
- Create an alias with a different destination and the box unticked → stored default
  unchanged.
- Two destinations with the box ticked → both stored and both prefilled again.
- ✕ → redirect back to the form, default gone, ✕ no longer shown.
- POST to the endpoint without a CSRF token → rejected.
- `composer check-format` and `composer lint` clean; `composer psalm` unchanged
  (only the pre-existing `PDO::MYSQL_ATTR_*` findings).

## Files

- `public/upgrade.php` — `upgrade_1855` migration (`admin_preferences`).
- `functions.inc.php` — `get_admin_pref()` / `set_admin_pref()` / `delete_admin_pref()`.
- `model/AliasHandler.php` — the `goto_default` field, prefill and storing on save.
- `templates/editform.tpl` — checkbox, stored value + ✕, label-switching JS.
- `public/alias-goto-default.php` — POST-only endpoint for deleting the stored value.
- `languages/de.lang`, `languages/en.lang` — new strings.


public/upgrade.php
```diff
@@ -2251,3 +2251,45 @@ function upgrade_1854()
     # add description after 'target_domain' field in alias_domain table
     _db_add_field('alias_domain', 'description', "varchar(255) {UTF-8} NOT NULL DEFAULT ''", 'target_domain');
 }
+
+function upgrade_1855_mysql()
+{
+    # per-admin preferences (key/value), e.g. the "goto" prefill setting for the add-alias form
+    $admin_preferences = table_by_key('admin_preferences');
```
DavidGoodwin: i don't think these upgrade() additions need splitting into per database ones ...
would a single upgrade_1855() { .. } not work with just : 

```
        CREATE TABLE {IF_NOT_EXISTS} $admin_preferences (
            `username` varchar(255) NOT NULL default '',
            `pref_key` varchar(64) NOT NULL default '',
            `pref_value` varchar(255) NOT NULL default '',
            PRIMARY KEY (`username`, `pref_key`)
        ) {COLLATE} COMMENT='Postfix Admin - Admin Preferences';
```

if one of the backends doens't like COMMENT='...' then perhaps just drop that?

templates/settings.tpl
```diff
@@ -0,0 +1,76 @@
+<form name="settings" method="post" action="" class="form-horizontal">
+    <div id="edit_form" class="panel panel-default">
+        <div class="panel-heading"><h4>{$PALANG.pMenu_settings}</h4></div>
+        <div class="panel-body">
+            <input class="flat" type="hidden" name="token" value="{$smarty.session.PFA_token|escape:"url"}"/>
```
DavidGoodwin: i think there's a CSRF helper function somewhere now to use ?

`{CSRF_Token}` might suffice.

DavidGoodwin (changes requested):

DavidGoodwin:
@mkocustos Sorry! I did comment on the PR last week (i think) but github's UI clearly confused me and I failed to submit the review, so you probably didn't see the comments.

Thank you for taking the time to contribute.

DavidGoodwin:
I'm wondering if there's an easier way to do this - e.g. having a tickbox or something underneath the destination field to 'remember me' for subsequent page loads (where the value is just stored in $_SESSION or something).

public/upgrade.php
mkocustos: Done — collapsed into a single `upgrade_1855()`.

I dropped both the inline `COMMENT='...'` and the backticks: `db_query_parsed()` substitutes the `{IF_NOT_EXISTS}`/`{COLLATE}` tokens per backend but doesn't rewrite identifier quoting or strip MySQL-only clauses, so backticks (PostgreSQL rejects them) and the inline COMMENT (MySQL-only) were what forced the per-backend split. The identifiers aren't reserved words in any of the three engines, so the DDL now runs as one function on MySQL/MariaDB, PostgreSQL and SQLite. Verified functionally on MariaDB and SQLite (migration reaches version 1855 and creates the table); PostgreSQL portability follows from the standard-SQL syntax + token substitution.

templates/settings.tpl
mkocustos: Done — switched to the `{CSRF_Token}` helper in the template and `CsrfToken::assertValid(safepost('CSRF_Token'))` in settings.php, matching the other POST handlers (sendmail.php, delete.php). Dropped the hand-rolled hidden field and $_SESSION['PFA_token'] comparison.

mkocustos:
Thanks for the review, @DavidGoodwin. Both points addressed and pushed:

- `upgrade.php`: single portable `upgrade_1855()` (no backticks, no inline COMMENT) — runs on MySQL/MariaDB, PostgreSQL and SQLite. Verified on MariaDB and SQLite.
- `settings.tpl` / `settings.php`: use the `{CSRF_Token}` helper and `CsrfToken::assertValid()`.

I also rebased the branch onto current `master` and added the Settings link to the dashboard action list (`main.tpl`) in addition to the top navigation, so it's reachable from the landing page too.

mkocustos:
> I'm wondering if there's an easier way to do this - e.g. having a tickbox or something underneath the destination field to 'remember me' for subsequent page loads (where the value is just stored in $_SESSION or something).
> 
Sorry, saw your message only afterwards. I'd like to take multi client use (laptop + mobile phone) into account, I'm not sure whether the remember option (bind to a session) would work in this case.

mkocustos:
Building on that idea — I think the *UI* half of your suggestion is a clear improvement, and I'm happy to adopt it: drop the separate Settings page (and its menu/dashboard entries, language strings, etc.) and instead put a small checkbox right under the destination field on the add-alias form, e.g. "remember this destination for new aliases". That removes most of the surface area of this PR.

The one part I'd like to keep is *where* the value is stored. If it lives in `$_SESSION`, it's per-browser and gone on session expiry, so it wouldn't carry across devices (my main use case is adding aliases from both a laptop and a phone). So my proposal is a hybrid:

- **UI:** inline checkbox under the destination field (your idea) — no settings page.
- **Storage:** persist the value per-admin in the `admin_preferences` table when the box is ticked, and prefill `goto` from it on the next add-alias load.

That keeps the footprint small while surviving session expiry and working across devices. It also lets me drop the `login`/`custom` modes and the mailbox-existence check, which were really only there to justify a dedicated settings screen.

The migration + CSRF fixes from the last round stay as they are regardless.

Does that direction sound good to you? If so I'll rework the PR accordingly (and the diff should shrink noticeably).

knofte:
Prefilling the alias "To" field, I think it's worth flagging that this field is a txtl type rendered as a <textarea> (an alias can forward to several addresses, one per line).

However, browsers only offer value-history/autocomplete on single-line <input> fields, not <textarea>s, so a browser-native "remember the last value" approach won't work here without changing the field type, which we wouldn't want (it'd break our multi-target alias entry).

The best way forward here @mkocustos if you really need prefilled alias, take a look at $CONF['alias_struct_hook'] which can already set it server-side from authentication_get_username() .

Good luck :)

knofte:
A "remember" tick, needs to be very specifically explained what it does and for what, so it is not seen as a "do I have to click remember for this one time alias to actually be saved" - or better renamed, or to be fair - not added, as this might be feature-creep.

(could rather change to input rather than textarea and have "expand" if more than one alias is added, and thus enable textarea )

mkocustos:
> A "remember" tick, needs to be very specifically explained what it does and for what, so it is not seen as a "do I have to click remember for this one time alias to actually be saved" - or better renamed, or to be fair - not added, as this might be feature-creep.
> 
> 
> 
> (could rather change to input rather than textarea and have "expand" if more than one alias is added, and thus enable textarea )

That argument would count for my first approach, wouldn't it? 

It's easy to understand and it is obvious for the user what is happening. However, the change is more invasive.

mkocustos:
Thanks both for the feedback — I think there might be a slight mix-up between two different mechanisms, so let me clarify:

Re: textarea/autocomplete (@knofte) — the prefill in both my original approach and David's checkbox suggestion doesn't rely on browser-native autocomplete/value-history at all. It works server-side: PFAHandler populates $this->values[$key] from $row['default'] in $struct before the form is rendered — that's the same generic mechanism editform.tpl already uses for every field type, goto included (it's a plain txtl/textarea field there, nothing special-cased). Since the default is set in PHP before Smarty renders the template, it works the same regardless of whether the field is an <input> or a <textarea> — there's no dependency on browser autofill behavior. So I don't think the textarea point is actually a blocker here, though happy to be corrected if I'm missing something.

Re: alias_struct_hook (@knofte) — that's a great pointer for a different use case, but I think it solves a different problem than what I'm after. alias_struct_hook is set in config.local.php, so it requires shell/config access to the server. My use case is domain admins who only have web UI access and want to set their own prefill preference without involving a sysadmin. So it's really "self-service per-admin preference" vs. "global sysadmin-defined rule" — both valid, but not overlapping.

Given that, I'd like to move ahead with the hybrid approach from my last comment (inline checkbox under the destination field, persisted per-admin in admin_preferences rather than session, to survive multi-device use).
Before I put more time into it though — @DavidGoodwin, does that direction sound right to you? I don't want to sink effort into an implementation if the overall approach isn't one you'd be open to merging.

DavidGoodwin:
@mkocustos My only concern is that it's a niche requirement (requiring persistence across browser sessions) and probably only something you'd ever use in your organisation. 

If others user's chipped in and expressed interest in the feature - then fine, I'd be happy to accept it.

( I'd have no problem with it if it was just a session based memory )

dsteinborn:
I would also like to have this feature implemented. Instead of using a catch-all email I'm creating custom aliases for many services I use. As I'm using multiple devices to create aliases (desktop, laptop, mobile) it would be great to save the default recipient session independently.

mkocustos:
Reworked along the lines we discussed — the Settings page is gone, and the whole thing is now a checkbox under the **To** field. The branch is rebased onto current `master`.

@DavidGoodwin — this is your inline-control suggestion, with the storage kept in `admin_preferences` rather than `$_SESSION` so it survives session expiry and works across devices, as @dsteinborn also asked for. Dropped with the settings page: the `login`/`custom` prefill modes, the mailbox-existence check, the menu entry and the dashboard tile.

@knofte — on the "remember" tick being ambiguous, I think that was a fair objection, so the label states the effect instead of naming it, and it is rewritten on click to describe the chosen state:

```
☐ Save this destination as the default "To" address for future aliases
☑ This destination will be saved as the default "To" address for future aliases
```

and, when a default already exists:

```
☐ Save this destination as the default "To" address for future aliases
☑ This destination will replace the saved default "To" address
   Currently saved: someone@example.org  ✕
```

The box is never pre-ticked, so a one-off alias to some other address can never silently overwrite a stored default — replacing one is always a deliberate click. The ✕ deletes it again straight away.

On your other two points: `alias_struct_hook` lives in `config.local.php` and so needs shell access, whereas this is meant for domain admins who only have the web UI — global sysadmin rule vs. self-service per-admin preference. And the prefill does not depend on browser value history: `PFAHandler` fills `$this->values[$key]` from `$row['default']` in `$struct` before Smarty renders, so `<textarea>` vs `<input>` makes no difference to it.

Two things worth flagging honestly, since they grew the diff back a little:

1. `editform.tpl` is now touched. Field descriptions are html-escaped on their way to the template, so the changing label and the ✕ cannot be smuggled in through the struct — but I hooked into the per-table special-case branches the template already provides rather than adding new ones.
2. Deleting the stored value has to work without creating an alias, so it gets a small POST-only endpoint (`public/alias-goto-default.php`, 49 lines, CSRF-checked exactly like `delete.php`, no menu entry, no template). If you would rather not have another file, the alternative is to make the deletion take effect only on the next alias save — I think that is worse, but happy to change it.

The label switching is ~14 lines of JS in the template's existing script block; without JS the feature still works, the label just stays on the unticked wording.

Migration and CSRF handling from the previous round are unchanged. Tested against a copy of a production database on MariaDB (and the migration on SQLite): storing, prefilling, replacing, multiple destinations, deleting via ✕, and a POST without a CSRF token being rejected. `composer check-format` and `composer lint` are clean, `psalm` unchanged apart from the pre-existing `PDO::MYSQL_ATTR_*` findings.

public/alias-goto-default.php
```diff
@@ -0,0 +1,49 @@
...
+ * LICENSE
+ * This source file is subject to the GPL license that is bundled with
+ * this package in the file LICENSE.TXT.
+ *
+ * Further details on the project are available at https://github.com/postfixadmin/postfixadmin
+ *
+ * @version $Id$
+ * @license GNU GPL v2 or later.
+ *
+ * File: alias-goto-default.php
```
cboltz: delete-pref.php or delete-preference.php would be a better filename for this file.

cboltz:
I admit that I'm not the biggest fan of this feature, but at least I now have an idea why it could be useful in some cases.

> @knofte — on the "remember" tick being ambiguous, I think that was a fair objection, so the label states the effect instead of naming it, and it is rewritten on click to describe the chosen state:
> 
> ```
> ☐ Save this destination as the default "To" address for future aliases
> ☑ This destination will be saved as the default "To" address for future aliases
> ```
> 
> and, when a default already exists:
> 
> ```
> ☐ Save this destination as the default "To" address for future aliases
> ☑ This destination will replace the saved default "To" address
>    Currently saved: someone@example.org  ✕
> ```

Why do I have the impression that you always search for the most complicated way when implementing something? ;-)

I don't like changing the label when ticking the checkbox - the "Save this destination ..." text combined with a ticked checkbox is more than obvious. It is also obvious that saving a new default will replace a previous default, no need to mention this ;-) (Besides that, changing text when ticking the checkbox in worst case means parts of the page "jump around".)

If a saved default exists, that default + the button to delete it should always be visible.

At the risk of disappointing you - no JS needed ;-)

> On your other two points: `alias_struct_hook` lives in `config.local.php` and so needs shell access, whereas this is meant for domain admins who only have the web UI — global sysadmin rule vs. self-service per-admin preference. And the prefill does not depend on browser value history: `PFAHandler` fills `$this->values[$key]` from `$row['default']` in `$struct` before Smarty renders, so `<textarea>` vs `<input>` makes no difference to it.

With some really evil tricks, I could imagine to handle this with alias_struct_hook even with the preference stored in the database. (I'm not saying such evil tricks are a good idea ;-)

mkocustos:
Thanks for the feedback so far — to avoid going back and forth again, here's what I'm planning for the next revision:

- Rename `alias-goto-default.php` to `delete-preference.php` (better reflects what the endpoint actually does)
- Remove the JS, per @cboltz's suggestion, and handle this with a plain server-side/no-JS approach instead

I'll admit it's been hard for me to converge on the "right" approach here, since there don't seem to be explicit coding, naming, or UI conventions documented anywhere that I could use as a guide — happy to be pointed to something I missed.

Before I put more time into this, does this direction work for all of you? Happy to adjust if not, but wanted to confirm before implementing a third rewrite.

DavidGoodwin:
@mkocustos sorry! As a sort of relevant summary, I feel the project has bounced along with random changes from me, cboltz or an occassional contributor, often with months between any activity. So there wasn't much need to agree/document a 'right' approach. 

With the advent of LLMs more recently, there now seems to be a new PR every other day, as many of the barriers to contributing are removed (there's no need for a human to read the existing code etc - you just tell claude to do X and wait).

I think most of your pain has been more of what functionality is actually required/is useful for others. I understand you have your own itch to scratch, but we're also trying to think of the future - avoid unnecessary scope creep and maintenance load.

I suspect we probably do need a claude.md and other files to try and help make life easier/quicker for LLMs and to steer them in the right direction, but that's a separate issue.

DavidGoodwin:
> Thanks for the feedback so far — to avoid going back and forth again, here's what I'm planning for the next revision:
> 
>     * Rename `alias-goto-default.php` to `delete-preference.php` (better reflects what the endpoint actually does)
> 
>     * Remove the JS, per @cboltz's suggestion, and handle this with a plain server-side/no-JS approach instead
> 
> 
> I'll admit it's been hard for me to converge on the "right" approach here, since there don't seem to be explicit coding, naming, or UI conventions documented anywhere that I could use as a guide — happy to be pointed to something I missed.
> 
> Before I put more time into this, does this direction work for all of you? Happy to adjust if not, but wanted to confirm before implementing a third rewrite.

Sounds good to me. Thank you for your efforts and coping with "us Luddites" :-)

mkocustos:
Pushed the revision we agreed on — smaller and with the JavaScript gone.

@cboltz — your points:

- **No more label switching / no JS.** The checkbox label is now static: *"Save this destination as the default 'To' address for future aliases"*. A ticked box next to that sentence is self-explanatory, and nothing on the page moves around any more. The `<script>` handling for it is removed entirely.
- **Stored default + delete button always visible.** When a default exists it is shown right under the checkbox as *"Currently saved: …"* with an inline ✕ to remove it — rendered server-side, no JS gating.
- I dropped the `_on` / `_on_replace` language strings that only existed to feed the old label switch.

@DavidGoodwin — as flagged, I also renamed the endpoint `alias-goto-default.php` → `delete-preference.php`, which better describes what it does (it deletes the stored admin preference; storing still happens in `AliasHandler::postSave()`).

The only remaining inline JS is the `onclick="return confirm(...)"` on the ✕ button, matching the existing convention in `list.tpl` / `list-virtual.tpl` for delete actions; it degrades gracefully (without JS the button just submits).

Net effect: the `editform.tpl` change is now +18 instead of +50, two fewer strings per language, no script block. Whole PR is 7 files / +222.

Tested against a copy of a production database on MariaDB: storing, prefilling, replacing, a one-off alias with the box unticked leaving the stored default alone, multiple destinations, deleting via ✕, a POST without a CSRF token being rejected, and a GET on the endpoint being a no-op redirect. `composer check-format` and `lint` are clean; `psalm` is unchanged apart from the pre-existing `PDO::MYSQL_ATTR_*` findings.

Thanks again for the patience with the back-and-forth.

PR #1067, opened 2026-07-02
Improve invalid CSRF token response

TrapoSAMA:
This ports the invalid CSRF token handling improvement from the 4.0 branch to master.\n\nChanges:\n- Keeps the HTTP 419 response for invalid CSRF tokens.\n- Replaces the raw hardcoded error with localized PALANG text.\n- Does not redirect to login.php, avoiding session destruction in multi-tab workflows.\n- Provides a simple Continue link back to main.php so normal authentication flow can decide whether login is needed.\n\nRelated to the 4.0 branch fix in PR #1055.


TrapoSAMA:
I found one more edge case while testing the invalid-token page locally: when the session has no CSRF_Tokens array, CsrfToken::assertValid() could emit a PHP warning before the friendly 419 response. I updated the guard to initialize the token list when it is missing or not an array, matching the existing generate() behavior. The latest GitHubBuild run is passing.

PR #1069, opened 2026-07-04
Add docs: custom fields via *_struct_hook (with domain backend_host example)

snathan100:
As discussed in #1068.

Adds `DOCUMENTS/CUSTOM_FIELDS.md`, a start-to-finish walkthrough for adding a custom field to a built-in entity using the `*_struct_hook` callbacks, no core patching required.

It covers:

- how the hook is picked up in `PFAHandler` and the full list of `*_struct_hook` config keys
- the two gotchas (the hook does not create the DB column; prefix custom columns per the naming policy)
- a worked example that adds a per-domain `backend_host` field with `PFAHandler::pacol()`, then wires it into Postfix `transport_maps` and Dovecot proxying so a front-end box can route each domain to its downstream mail server
- a short note on why a dedicated hostname column can be nicer than reusing `transport` when the same value also feeds Dovecot's proxy

Happy to adjust scope, wording, or the filename. In particular I'm not sure whether you'd prefer this framed as a general "custom fields" guide (as written) or narrowed to the backend-routing use case.

### DIFF
--- a/DOCUMENTS/CUSTOM_FIELDS.md
+++ b/DOCUMENTS/CUSTOM_FIELDS.md
@@ -7,9 +7,11 @@ which let you add, change, or remove fields on a Handler's `$struct` at runtime.
 
 The handy part is that the Handler builds both its SQL and its edit form
 straight from `$struct`, so a field you bolt on through the hook gets full add,
-edit, and list support in the web UI without any extra work.
+edit, and list support in the web UI without any extra work. It also gets picked
+up automatically by `postfixadmin-cli`, so the same field is manageable from the
+command line, not just the browser.
 
-The worked example below adds a per-domain `backend_host` field and then wires
+The worked example below adds a per-domain `x_backend_host` field and then wires
 it into Postfix (`transport_maps`) and Dovecot (proxy). The end result is a
 front-end PostfixAdmin box that routes each domain to whichever downstream
 server actually holds its mailboxes. The same recipe works for any custom
@@ -51,16 +53,16 @@ Don't forget to return it.
 1. The hook doesn't touch the database. Adding a field to `$struct` tells
    PostfixAdmin how to *handle* the column, but it won't *create* it for you.
    You add the column yourself with `ALTER TABLE`.
-2. Prefix your custom columns so they don't collide with anything a future
-   PostfixAdmin release might add. There's a naming policy for exactly this:
-   <https://sourceforge.net/p/postfixadmin/wiki/Custom_fields/>. This guide
-   sticks with plain `backend_host` because it reads better, but for real
-   deployments something like `x_backend_host` is the safer bet.
+2. Prefix your custom columns with `x_` so they don't collide with anything a
+   future PostfixAdmin release might add. There's a naming policy for exactly
+   this: <https://sourceforge.net/p/postfixadmin/wiki/Custom_fields/>. That's
+   why the example below uses `x_backend_host` rather than a bare
+   `backend_host`.
 
 ## Step 1: add the database column
 
 ```sql
-ALTER TABLE domain ADD COLUMN backend_host VARCHAR(255) NULL DEFAULT 'backend1.example.com';
+ALTER TABLE domain ADD COLUMN x_backend_host VARCHAR(255) NULL DEFAULT 'backend1.example.com';
 ```
 
 ## Step 2: surface the field with a hook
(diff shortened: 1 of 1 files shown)


DOCUMENTS/CUSTOM_FIELDS.md
```diff
@@ -0,0 +1,157 @@
...
+won't wipe it out. Point the config key at a function and define that function.
+The tidiest way to build the column definition is `PFAHandler::pacol()`, since
+that's how the core defines its own fields:
+
+```php
+$CONF['domain_struct_hook'] = 'domain_struct_hook';
+
+function domain_struct_hook($struct) {
+    $struct['backend_host'] = PFAHandler::pacol(
+        1,                 // editable
```
DavidGoodwin: this could be rewritten to be .... 
```php
$struct['backend_host'] = PFAHandler::pacol(  
   allow_editing: 1,
   display_in_form: 1,
   display_in_list: 1,
   type: 'text',
   PALANG_label: 'Backend Host',
   PALANG_desc: 'Downstream .... ' 
   default: 'backend1.example.com'
)
```

Is using PHP's named parameters easier to understand / cleaner?

DOCUMENTS/CUSTOM_FIELDS.md
```diff
@@ -0,0 +1,157 @@
...
+Don't forget to return it.
+
+## Two things to know before you start
+
+1. The hook doesn't touch the database. Adding a field to `$struct` tells
+   PostfixAdmin how to *handle* the column, but it won't *create* it for you.
+   You add the column yourself with `ALTER TABLE`.
+2. Prefix your custom columns so they don't collide with anything a future
+   PostfixAdmin release might add. There's a naming policy for exactly this:
+   <https://sourceforge.net/p/postfixadmin/wiki/Custom_fields/>. This guide
```
DavidGoodwin: I didn't know this wiki page existed, we should really copy it into git ...

DavidGoodwin (approved):

DOCUMENTS/CUSTOM_FIELDS.md
```diff
@@ -0,0 +1,157 @@
+# Custom Fields with `*_struct_hook`
+
+This is a walkthrough for adding your own field to one of PostfixAdmin's
+built-in entities (domains, mailboxes, aliases, admins, and so on) without
+patching any PostfixAdmin code. It uses the `*_struct_hook` config callbacks,
+which let you add, change, or remove fields on a Handler's `$struct` at runtime.
+
+The handy part is that the Handler builds both its SQL and its edit form
+straight from `$struct`, so a field you bolt on through the hook gets full add,
+edit, and list support in the web UI without any extra work.
```
cboltz: It also automagically gets supported in `postfixadmin-cli`, not only in the web UI.

DOCUMENTS/CUSTOM_FIELDS.md
cboltz: Agreed.

In this guide, it would be a good idea to at least mention "prefix custom field names with `x_`. And it would be even better if the guide would actually use `x_backend_host` ;-)

DOCUMENTS/CUSTOM_FIELDS.md
cboltz: It probably depends ;-)
- if someone want to add a single field, comments or named parameters make sense (not sure which one is better)
- if someone wants to add multiple fields, then the "table-style" layout as in the *Handler classes + the comment header is more helpful

DOCUMENTS/CUSTOM_FIELDS.md
```diff
@@ -0,0 +1,157 @@
...
+$CONF['domain_struct_hook'] = 'domain_struct_hook';
+
+function domain_struct_hook($struct) {
+    $struct['backend_host'] = PFAHandler::pacol(
+        1,                 // editable
+        1,                 // display_in_form
+        1,                 // display_in_list
+        'text',            // type
+        'Backend Host',    // label (or a PALANG key)
+        'Downstream mail server that hosts this domain', // description
```
cboltz: Label and description should point to a (custom or existing) string in $PALANG, see `$CONF['language_hook'] = '';` for custom strings.

DOCUMENTS/CUSTOM_FIELDS.md
```diff
@@ -0,0 +1,157 @@
...
+side. If you only need the Postfix routing and don't care about Dovecot, then
+reusing `transport` (maybe with a bit of SQL `CONCAT`/`SUBSTRING` to reshape it)
+works just as well.
+
+## The short version
+
+- `*_struct_hook` lets you add, change, or drop Handler fields from config
+  alone. No core edits, no fork.
+- Add the DB column yourself. The hook only handles it, it doesn't create it.
+- Prefix your custom columns per the naming policy.
```
cboltz: Pointing to the naming policy is fine, but I'd mention the `x_` prefix here nevertheless.

snathan100:
Thanks for the quick review, both of you. I've pushed a commit applying all three suggestions:

- `pacol()` example now uses PHP 8 named arguments (much easier to read than counting positional params, good call)
- switched the whole example to `x_backend_host` so it actually follows the naming policy instead of just mentioning it
- added a note that struct-hook fields are picked up by `postfixadmin-cli` too, not only the web UI

Let me know if there's anything else you'd like tweaked.

DOCUMENTS/CUSTOM_FIELDS.md
cboltz: (Resolved for CUSTOM_FIELDS.md, just keeping open as a reminder to copy over the wiki page.)

cboltz:
Thanks for the update! It already looks better, but you seem to have missed my comment to use language_hook, and to mention the x_ prefix in the "short version" section.

snathan100:
Good catches, thanks — both fixed in the latest commit:

- **`language_hook`**: `PALANG_label`/`PALANG_desc` now point to `x_`-prefixed `$PALANG` keys, and I added a `$CONF['language_hook']` example that defines the strings (with a note on covering all languages/the default). Also mentioned that literal text still renders as a fallback, but the hook is the clean way.
- **`x_` prefix in the summary**: the "short version" now spells out the `x_` prefix explicitly (for both columns and custom `$PALANG` keys), plus a line pointing to `language_hook` for labels.

Noted on keeping the naming-policy thread open as a reminder to copy the wiki page over — makes sense.

DavidGoodwin:
perhaps just add the sf.net docs into the above?  I think this is all that's required ?


# Custom changes (??)

From time to time we get requests for adding some fields in PostfixAdmin. Many of them are specific for one user and therefore don't end up in the official code.

This page describes what you can do yourself.

## Policy for database changes

This policy is important to avoid that your custom fields conflict with any fields, tables or indexes we might add in future versions of PostfixAdmin.

## Additional fields in existing tables

 * you can use any fieldname starting with "x_"
 * you can create indexes if needed - also prefix them with x_ (or tablename_x_ for PostgreSQL)
 * avoid creating unique indexes and foreign keys if possible (they could make scheme changes difficult)

## Additional tables

 * use a table name starting with "x_"
 * within that table, you can use any field and index names you want
 * you might want to have the fields "created", "modified" and "active" (which exist in most PostfixAdmin tables)
 * you may also create unique keys inside your tables
 * if you use PostgreSQL, names for indexes and unique keys must start with "x_" (better: "x_tablename_")
 * avoid foreign keys to PostfixAdmin's default tables if possible.

snathan100:
Good idea. I've inlined that policy into the doc (new "Naming policy for custom fields and tables" section) so it lives in the repo instead of only linking out to sf.net, and pointed the earlier reference at it.

One small addition: since most installs are MySQL/MariaDB, I added a short note explaining that the `x_` prefix on index/key *names* is mainly a PostgreSQL thing (PG names them per-schema, so they can clash; MySQL/MariaDB scope index names per table). Indexes added to an existing core table are still best prefixed on any DB. Happy to trim that if you'd rather keep it verbatim.

DavidGoodwin:
thanks @snathan100 !

PR #1072, opened 2026-07-10
Add source install upgrade helper

TrapoSAMA:
## Summary
- Add scripts/installto.sh, a PHP CLI helper for updating source/archive installations from an unpacked newer release tree.
- Preserve local configuration files and theme assets referenced from config.local.php before replacing release files.
- Run install.sh and then apply the existing database upgrade through public/upgrade.php so the DB step is not forgotten.
- Document usage in INSTALL.md, including --dry-run and the rsync requirement.

## Usage
From the unpacked new release tree:

scripts/installto.sh --dry-run /srv/postfixadmin
scripts/installto.sh /srv/postfixadmin

Run it as root or as the deployment user that owns the installation. The web server user does not need write access to the code tree; templates_c remains the writable runtime directory.

## Validation
- php -l scripts/installto.sh
- php scripts/installto.sh --help
- php scripts/installto.sh --dry-run -y /var/www/postfixadmin
- git diff --check

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -75,6 +75,29 @@ Which will
 * download dependent PHP libraries (`composer install`)
 * create a templates_c directory if one does not exist.
 
+## 1a. Upgrade an existing source installation
+
+If you installed PostfixAdmin from the source archive (not from a distribution
+package), you can update an existing installation from an unpacked newer release
+with:
+
+```bash
+scripts/installto.sh /srv/postfixadmin
+```
+
+Run the command as root or as the deployment user that owns the installation
+tree. Use `scripts/installto.sh --dry-run /srv/postfixadmin` to preview the
+steps first. The helper is a PHP CLI script and the real update requires
+`rsync`.
+
+The web server user does not need write access to PostfixAdmin files; only
+`templates_c/` should be writable by the web server.
+
+The script copies the new release into the target directory, preserves
+`config.local.php`, preserves locally referenced theme assets such as
+`theme_logo`, runs `install.sh` in the target, and then applies the database
+upgrade through `upgrade.php`.
+
 ## 2.Setup Web Server
 
 Assuming /var/www/html is where your webserver reads from, either create a symlink:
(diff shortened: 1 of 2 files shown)


DavidGoodwin (approved):

DavidGoodwin:
while the idea has some merit, I don't really see much need for it. 

Many people will use packages provided by a distribution - which will handle all of this transparently. 

For those installing from source, an upgrade may be something like : 

```shell
wget https://github.com/postfixadmin/postfixadmin/releases/download/v4.0.5/postfixadmin-v4.0.5-php84.tar.gz
tar -zxf postfixadmin-v4.0.5-php84.tar.gz
mv postfixadmin postfixadmin.old
mv postfixadmin-v4.0.5 postfixadmin
cp postfixadmin.old/config.local.php postfixadmin/config.local.php
cd postfixadmin
bash install.sh
```

then just visit https://postfixadmin.whatever/setup.php so upgrade.php runs.

but they could be, like me, just needing to do : 

```shell
cd /var/www/postfixadmin
git fetch
git checkout v4.0.5
bash install.sh
```

and then visit setup.php in a browser....

With this PR/approach, I'd need to download a new .tar.gz anyway, explode it, and then figure out how to run a new shell script (the `installto.sh /var/www/postfixadmin`). 

I'm not sure that's much of a gain?

TrapoSAMA:
Thanks, that's a fair point. The main case I was trying to cover was archive-based installations with several local files or assets, while avoiding stale files and a forgotten database upgrade. However, I agree that the current helper adds a fair amount of machinery to a workflow that is normally only a few straightforward commands.

Keeping that simple is more valuable than adding an optional abstraction here. I'll leave it to you whether this PR should be closed, or whether a much smaller documentation-only clarification would still be useful.

DavidGoodwin:
I'd rather documentation updates (if they're needed).

TrapoSAMA:
Thanks, understood. I agree that documentation-only is the better direction if a concrete gap is identified. I'll close this implementation PR, and I won't add a documentation follow-up unless there is something specific in the existing upgrade guidance that needs clarification.

PR #1078, opened 2026-07-19
Unify user and admin forwarding editor

TrapoSAMA:
## Context

This proposal is based on the idea from #677 (`Forward admin users merge`): use one forwarding flow for both administrators and mailbox users, and remove the duplicated legacy user editor.

Rather than porting the old controller introduced there, this rebuilds the idea on the current `master` architecture. The shared `public/edit.php` entry point and `AliasHandler` provide the common implementation, which also resolves the remaining #677 TODO of using `templates/editform.tpl`.

## Process

1. Reused the generic handler/edit-form pipeline for both roles without adding role-specific capability flags to the generic controllers.
2. Linked the user menu and dashboard directly to the shared editor, deriving the record ID from the authenticated session and URL-escaping it in Smarty.
3. Made `AliasHandler` the authorization boundary: users can edit only their own record, and user access still respects `edit_alias`.
4. Rejected user-mode alias creation in `validate_new_id()`, exposed only the user's own alias in list mode, hid administrative list fields through `display_in_list`, and disabled deletion through `_can_delete`.
5. Kept administrative fields read-only for users while allowing only `goto` and `goto_mailbox` changes.
6. Removed the legacy `public/users/edit-alias.php` controller and its template instead of retaining compatibility or redirect wrappers. The shared form does not add a special Cancel action.
7. Preserved mailbox-local delivery and the vacation target while changing external forwarding destinations.
8. Kept the related Spanish translation improvements in this PR.

## Security and behavior

- The session-derived URL is navigation only; `AliasHandler` independently enforces ownership, so changing `edit=` cannot grant access to another mailbox.
- Submitted `address`, `description`, and `active` values cannot modify administrative fields.
- User-mode list queries are restricted by `PFAHandler::$user_field` and return only the authenticated user's mailbox alias.
- User alias creation and deletion are rejected by handler-level mechanisms.
- Domain-admin access still respects `alias_control_admin`.
- Vacation aliases are managed only by the vacation flow, cannot be forged through the forwarding form, and are preserved when already active.
- Duplicate forwarding destinations are removed.
- Forward-and-store, forward-only, and local-delivery-only configurations remain supported.

## Validation

- `AliasHandlerTest`: 16 tests, 744 assertions.
- Psalm: no errors.
- Parallel PHP lint: clean.
- GNU `patch -p1 --dry-run`: clean, and the applied tree matches the PR head.
- The XAMPP visual copy was refreshed with the reviewed implementation using PHP 8.4.

Original proposal: https://github.com/postfixadmin/postfixadmin/pull/677


DavidGoodwin:
@cboltz you might be better placed to review / accept this than me

public/users/edit-forward.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Redirect users to the shared AliasHandler forwarding form.
+ */
+
+require_once(__DIR__ . '/../common.php');
+
+authentication_require_role('user');
+
+if (!Config::bool('edit_alias')) {
+    header('Location: main.php');
```
cboltz: This check is usesess - nobody stops users to manually open `../edit.php` in their browser even if edit-forward.php refuses to redirect them there.

The permission check needs to be done directly in AliasHandler.

This also means the whole edit-forward.php file is superfluous - just change menu.conf and users_main.tpl to directly link to ../edit.php?table=alias (menu.conf doesn't support adding `&edit=$username`, so it should probably be taken from the login session. Actually edit.php already does that if I get line 165 right.)

public/delete.php
```diff
@@ -53,6 +53,11 @@
     if (empty($formconf['user_hardcoded_field'])) {
         throw new \InvalidArgumentException($handlerclass . ' is not available for users');
     }
+    if (isset($formconf['user_can_delete']) && !$formconf['user_can_delete']) {
+        flash_error(Config::lang_f('edit_not_allowed', $table));
+        header("Location: " . $formconf['listview']);
+        exit;
+    }
```
cboltz: Please adjust the `_can_delete` "field" in init_struct() instead of adding another check here.

public/edit.php
```diff
@@ -60,6 +60,11 @@
     if (empty($formconf['user_hardcoded_field'])) {
         throw new \InvalidArgumentException($handlerclass . ' is not available for users');
     }
+    if ($new && isset($formconf['user_can_create']) && !$formconf['user_can_create']) {
+        flash_error(Config::lang_f('edit_not_allowed', $table));
+        header("Location: " . $formconf['listview']);
+        exit;
+    }
```
cboltz: Better do this check via `validate_new_id()` - for aliases, just set an error message and return false if a user/non-admin is logged in.

(Setting `$msg['can_create']` in AliasHandler would be even better - but so far this variable only hides the "Add $whatever" button in list view. If you want to change/extend this, please do it as a separate PR, and see `git grep can_create` for the current (very limited) usage.)

DOCUMENTS/HANDLER_CLASSES.md
```diff
@@ -167,6 +167,10 @@ public function webformConfig()
         'listview' => 'list.php?table=foo',   // redirect target after edit
         'early_init' => 0,                    // 1 = call init() before set()
         'user_hardcoded_field' => 'username', // required if users can access
+        'user_can_create' => true,            // optional per-entry-point restriction
+        'user_can_list' => true,
+        'user_can_delete' => true,
```
cboltz: See the comments in the files using these user_can_* variables - IMHO they shouldn't be part of webformConfig(), and instead be checked directly in the relevant functions in *Handler.

(cancelview makes more sense, but I wonder if we really want/need a cancel button except for historical reasons - or if just dropping that button would also be ok.)

model/AliasHandler.php
```diff
@@ -124,17 +138,25 @@ public function webformConfig()
...
             'required_role' => 'admin',
-            'listview'      => 'list-virtual.php',
+            'listview'      => $listview,
             'early_init'    => 0,
             'prefill'       => array('domain'),
+            'user_hardcoded_field' => 'address',
+            'user_can_create' => false,
+            'user_can_list' => false,
+            'user_can_delete' => false,
+            'cancelview' => $this->is_admin ? '' : 'users/main.php',
```
cboltz: Assuming we keep the cancel button (see the comment for HANDLER_CLASSES.md), please use `$listview` instead of repeating `users/main.php`. (Doesn't change anything now, but makes long-term maintenance easier.)

public/users/edit-alias.php
```diff
@@ -1,149 +1,7 @@
...
-    } else {
-        $smarty->assign('forward_and_store', '');
-        $smarty->assign('forward_only', ' checked="checked"');
-    }
-
-    $smarty->display('index.tpl');
-}
-
-/* vim: set expandtab softtabstop=4 tabstop=4 shiftwidth=4: */
+require_once(__DIR__ . '/edit-forward.php');
```
cboltz: I slightly doubt that someone keeps edit-alias.php in their bookmarks ;-) (especially because it's behind the login, and would redirect to login first) - but if you really want to keep the old file, please make it a redirect so that people at least get the up-to-date URL for their bookmarks.

That said - I'd tend to just delete the file.

public/list.php
```diff
@@ -65,6 +65,11 @@
     if (empty($formconf['user_hardcoded_field'])) {
         throw new InvalidArgumentException($handlerclass . ' is not available for users');
     }
+    if (isset($formconf['user_can_list']) && !$formconf['user_can_list']) {
+        flash_error(Config::lang_f('edit_not_allowed', $table));
+        header("Location: " . $formconf['listview']);
+        exit;
+    }
```
cboltz: I'm not a fan of adding this check here.

If a user opens `list.php?table=alias`, they'll see a list with exactly one item - their own alias. That's also what they'll see in the edit form ;-)

Better adjust the `display_in_list` setting in initStruct if you want to hide specific fields from non-admins.

tests/AliasHandlerTest.php
```diff
@@ -32,6 +33,216 @@ public function testBasic()
...
+        $this->assertSame(0, $struct['active']['display_in_form']);
+
+        $this->assertTrue($handler->set([
+            'address' => 'bob@example.com',
+            'description' => 'changed by user',
+            'goto' => ['forward@example.net'],
+            'goto_mailbox' => 1,
+            'active' => 0,
+        ]));
+        $this->assertTrue($handler->save(), json_encode($handler->errormsg));
```
cboltz, line 70: This test is confusing since most fields (except goto and goto_mailbox) are read-only for non-admins (and will silently keep their old values).

Please add a comment saying so.

TrapoSAMA:
Thanks for the detailed review — the suggested direction makes sense and is more consistent with the existing handler architecture.

I’ll rework the proposal as one coherent follow-up rather than push partial fixes. The next iteration will move the remaining permission decisions into `AliasHandler`, remove the generic `user_can_*` additions, link directly to the shared editor, use the existing `_can_delete` / `validate_new_id()` / `display_in_list` mechanisms, reconsider whether the cancel and compatibility routes are needed, and clarify the read-only-field behavior in the test.

I’ll respond to the individual threads once the code and regression tests have been updated together.

public/users/edit-forward.php
TrapoSAMA: Implemented in 7de5d631. The wrapper is gone: both user navigation links now open the shared editor directly with the URL-escaped username from the authenticated session. AliasHandler remains the authorization boundary and rejects attempts to edit another address.

public/delete.php
TrapoSAMA: Implemented in 7de5d631. The generic delete.php check was removed, and AliasHandler now selects 0 as _can_delete in user mode.

public/edit.php
TrapoSAMA: Implemented in 7de5d631. The generic edit.php capability check was removed; AliasHandler::validate_new_id() now rejects alias creation in user mode. The existing can_create message flag is only used to hide the list action.

DOCUMENTS/HANDLER_CLASSES.md
TrapoSAMA: Implemented in 7de5d631. The user_can_* and cancelview additions were removed from webformConfig(), the generic entry points and the documentation. Permission decisions now live in AliasHandler, and the extra Cancel button was dropped.

model/AliasHandler.php
TrapoSAMA: The Cancel button was removed in 7de5d631, so no separate cancel target is needed.

public/users/edit-alias.php
TrapoSAMA: Implemented in 7de5d631. The legacy edit-alias.php entry point and its unused template were deleted; navigation now goes directly to the shared editor.

public/list.php
TrapoSAMA: Implemented in 7de5d631. The generic list.php denial was removed. User-mode AliasHandler lists only the authenticated user mailbox alias via user_field scoping, while administrative columns are hidden through display_in_list and deletion remains disabled through _can_delete.

tests/AliasHandlerTest.php
TrapoSAMA: Implemented in 7de5d631. The test now explicitly explains that fields other than goto and goto_mailbox are read-only for users and silently retain their stored values.

model/AliasHandler.php
```diff
@@ -144,6 +171,13 @@ public function webformConfig()
      */
     public function init(string $id): bool
     {
+        if (!$this->is_admin && !$this->new) {
```
cboltz: Just wondering - what's the reason for adding the "not new" restriction here?

If there isn't a specific reason, it can't hurt to only check for "non-admin" and run the strict test (even if a create attempt should already be blocked by validate_new_id).

TrapoSAMA: Agreed - there was no specific reason to exclude new records from this check. I removed the `!$this->new` condition in 31721184, so every non-admin request now goes through the strict ownership/edit_alias check. `validate_new_id()` remains the second barrier against user alias creation. The focused AliasHandler test suite still passes: 16 tests, 744 assertions.

TrapoSAMA:
Separately, if this unified forwarding editor change is accepted for `master`, would you also consider it for the `postfixadmin_4.0` branch? I can prepare a separate 4.0-compatible proposal so the backport can be reviewed independently.

model/AliasHandler.php
TrapoSAMA: Follow-up in d4f608a8: the user-creation regression test now calls init() with the authenticated address itself. This deliberately passes the strict ownership comparison and verifies that validate_new_id() independently rejects user alias creation. All GitHub checks pass.

DavidGoodwin:
> Separately, if this unified forwarding editor change is accepted for `master`, would you also consider it for the `postfixadmin_4.0` branch? I can prepare a separate 4.0-compatible proposal so the backport can be reviewed independently.

I'd rather not. We're not trying to actively develop 4.0 - it's in a sort of maintenance only mode. If there are bugs to fix then fine, but I don't think that's the case here.

cboltz:
Agreed, this PR doesn't qualify as a bugfix, even if "use MailboxHandler everywhere (and get rid of separate code)" is a big improvement.

For master, your PR looks good, and I'll merge it. Thanks for the PR and for fixing all the things I complained about ;-)

PR #1080, opened 2026-07-19
Harden optional mailbox hook scripts

TrapoSAMA:
## Summary

- backport only the generic hardening of the existing optional Maildir and Dovecot hook examples
- validate the documented hook argument contracts before constructing filesystem paths
- quote filesystem operations, replace legacy shell constructs and propagate command failures
- make domain and mailbox deletion safely retryable when their target is already absent
- update the Dovecot password example for Dovecot 2.3.19 or newer, with a Dovecot 2.4 setting note

This PR deliberately does **not** include the optional Rspamd DKIM scripts proposed for master in #1079.

## Why

The existing deletion examples accept missing arguments, allowing an incorrect or direct invocation to derive the Maildir base directory itself as the move source. They also rely on weak double-dot checks, unquoted paths and incomplete exit-status handling. The password hook keeps the old Dovecot invocation active.

## Impact

The hooks remain optional and disabled by default. Administrators must review `basedir`, `trashbase` and the service account used by sudo. The mail-crypt password example now requires Dovecot 2.3.19 or newer.

## Validation

- ShellCheck at style severity
- `sh -n` and `bash -n` syntax validation
- isolated functional tests for creation, deletion, invalid traversal input and retryable deletion
- NUL-delimited password-hook tests with simulated `openssl` and `doveadm`
- GNU `patch -p1 --dry-run` and full application from a clean `origin/postfixadmin_4.0` worktree
- exact tree comparison between the applied patch and this commit

## Rspamd follow-up question

The optional, privilege-separated Rspamd DKIM helpers and their deployment documentation are included in the master proposal #1079 as a separate second commit. Would maintainers also like that optional commit backported to `postfixadmin_4.0`, or should the Rspamd example remain master-only?

### DIFF
--- a/ADDITIONS/README.md
+++ b/ADDITIONS/README.md
@@ -48,22 +48,33 @@ See also :  https://github.com/postfixadmin/postfixadmin/tree/master/ADDITIONS/s
 by George Vieira <george at citadelcomputer dot com dot au>
 Deletes all unused mailboxes
 
-## Example mailbox / domain scripts for Postfixadmin
+## Example mailbox / domain scripts for PostfixAdmin
 
 - postfixadmin-mailbox-postcreation.sh
 - postfixadmin-mailbox-postdeletion.sh
 - postfixadmin-domain-postdeletion.sh
+
 by Troels Arvin <troels@arvin.dk>
 
-Examples of scripts relevant to the optional 
+Examples relevant to the optional `mailbox_postcreation_script`,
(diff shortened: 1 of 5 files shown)


DavidGoodwin:
Thank you! 

(I'm happy adding functionality to 4.0, especially if it's just something that's optional ... so if you want to do a Rspamd PR then I'll probably merge it)

Issue #971, opened 2026-01-25
Databse upgrades are inconsistent between 3.3 and 4.0


csware:
If you first upgrade to Postfixadmin 3.3.16 and then to 4.0, then not all database upgrades will be executes (i.e., 1846 does not update fetchmail table on 3.3 branch, but on 4.0 branch, cf. 08ce130d58aee88afee903fe559766f81651674e).

csware:
This seems as a blocker to me preventing me to upgrade to 4.0 now.

TrapoSAMA:
Thanks for reporting this. I plan to propose a focused fix for the 4.x branch by adding a new MySQL/MariaDB upgrade step, rather than changing `upgrade_1846_mysql()`. The goal will be to reconcile installations that first reached database version 1847 through PostfixAdmin 3.3.x and then upgraded to 4.x. Please keep an eye on this issue; I will link the proposed change here once it is ready for review.

TrapoSAMA:
The proposed fixes are now available for review: #1088 targets the `postfixadmin_4.0` branch, and #1089 targets `master`. Both are draft PRs while the initial CI/review cycle runs.

PR #1088, opened 2026-07-23
Reconcile 3.3 MySQL upgrades on 4.x

TrapoSAMA:
## Summary

- add `upgrade_1852_mysql()` to reconcile MySQL/MariaDB schemas that previously passed through the PostfixAdmin 3.3 upgrade path
- use `quota2.username = varchar(255)` as the marker for that path and replay the complete `upgrade_1846_mysql()` only when needed
- bump the minimum database version to 1852 for the next 4.x release

## Root cause

The 3.3 branch contains a partial `upgrade_1846_mysql()` and then records database version 1847 after widening `quota2.username`. When that database is later upgraded to 4.x, the more complete 4.x implementation of migration 1846 is skipped because the recorded version is already higher.

For a direct 4.x installation, `quota2.username` is already `varchar(100)`. In that case the new migration does not alter the schema and only advances the recorded database version.

Addresses #971.

## Validation

- PHP 8.4 syntax and parallel lint checks
- PHP-CS-Fixer dry run
- MariaDB 10.4 integration test reproducing the affected schema at database version 1851
- verified `quota2.username` normalization from `varchar(255)` to `varchar(100)`
- verified all affected column and table collations match a clean 4.x reference schema
- verified `vacation_notification_pkey` remains present
- verified the already-correct `varchar(100)` path is a schema no-op

### DIFF
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -16,7 +16,7 @@
  */
 
 
-$min_db_version = 1851;  # update (at least) before a release with the latest function number in upgrade.php
+$min_db_version = 1852;  # update (at least) before a release with the latest function number in upgrade.php
 
 
 /**
--- a/public/upgrade.php
+++ b/public/upgrade.php
@@ -2412,3 +2412,34 @@ function upgrade_1851_mysql()
         )
     ");
 }
+
+/**
+ * Reconcile MySQL schemas upgraded through the PostfixAdmin 3.3 branch.
+ *
+ * PostfixAdmin 3.3 used upgrade_1847_mysql() to widen quota2.username to
+ * varchar(255). A later upgrade to 4.x therefore skips the more complete
+ * upgrade_1846_mysql() in this branch because the recorded database version
+ * is already 1847. The widened quota2 column is used as the marker for that
+ * upgrade path; a direct 4.x upgrade leaves it at varchar(100).
+ *
+ * See https://github.com/postfixadmin/postfixadmin/issues/971
+ */
+function upgrade_1852_mysql()
+{
+    $quota2 = table_by_key('quota2');
+    $column = db_query_one(
+        "SELECT CHARACTER_MAXIMUM_LENGTH AS character_maximum_length
+        FROM information_schema.COLUMNS
+        WHERE TABLE_SCHEMA = DATABASE()
+          AND TABLE_NAME = :table
+          AND COLUMN_NAME = 'username'",
+        ['table' => trim($quota2, '`')]
+    );
+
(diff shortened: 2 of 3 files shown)


DavidGoodwin:
Thanks @TrapoSAMA

knofte:
There is one important problem/issue here though..

What it does is sound in general like... use quota2.username = varchar(255) as a marker that a schema came through the 3.3 path (where the 4.x upgrade_1846 got skipped), and if so, replay the full upgrade_1846_mysql(). Marker logic is reasonable; a clean 4.x install is varchar(100) so it's a correct no-op there.

However, the risk: replaying the entire 1846 is a blunt instrument, and 1846 contains two non-idempotent statements:
  ALTER TABLE …vacation_notification DROP FOREIGN KEY vacation_notification_pkey
  …
  ALTER TABLE …vacation_notification ADD CONSTRAINT vacation_notification_pkey FOREIGN KEY …
 failure mode is real, db_query() throws by default ($ignore_errors=false), and _do_upgrade() calls each upgrade_*() with no try/catch, so if any replayed statement errors, the whole upgrade halts mid-migration, leaving a half-done schema which is worse than before.
 
The DROP fails if the FK is absent, while the ADD fails if it's already present. :+1: 

Worth an extra view here @TrapoSAMA  :)

TrapoSAMA:
﻿Thanks @knofte ? good catch. You were right that replaying the migration was not safe enough to retry after a partial DDL failure.

I've updated this PR to:

- check `information_schema` before dropping or adding `vacation_notification_pkey`
- keep the `quota2.username = varchar(255)` path marker until every other reconciliation statement has succeeded
- cover the normal path, the already-correct no-op path, a forced foreign-key restoration failure, and the successful retry with MariaDB tests

The marker ordering exposed an additional issue: previously, a failure after normalizing it could make the next run treat an incomplete reconciliation as a no-op. The new ordering avoids that.

Because the same pattern had already been merged into `master` through #1089, I also opened #1092 with the equivalent retry-safety correction; all of its CI checks are passing.

Thanks again for the careful review ? this materially improved the safety of the migration.

DavidGoodwin:
@TrapoSAMA is this PR still needed?

TrapoSAMA:
﻿Yes, it is still needed for `postfixadmin_4.0`. This PR contains both the original reconciliation merged into `master` through #1089 and the retry-safety correction merged through #1092. Those PRs only updated `master`, so without #1088 the 4.0 branch would not receive either part of the complete fix.

Thanks for checking.

Issue #972, opened 2026-01-25
Postfixadmin should make more use of foreign keys


csware:
Missing foreign keys are (at least for MySQL/MariaDB):

* fetchmail.destuser -> mailbox->username
* quota.username -> mailbox->username
* quota2.username -> mailbox->username
* vacation.email -> mailbox->username
* domain_admins.domain -> domains.domain
* domain_admins.username -> admins.username (I'm unsure whether this is already there)

This also requires some cleanup steps beforehand such as `DELETE FROM $vaction where email not in (SELECT username from $mailbox)` etc., because otherwise creating foreign keys may fail.

DavidGoodwin:
> I'm also not a big fan of doing automatic "cleanups" using DELETE queries - in worst case, "too much" gets deleted. So if we really add more foreign keys, I'd prefer to show the problematic rows to the user instead of blindly deleting them. (In theory there shouldn't be such rows, unless the user has manually modified the database in an incomplete way.)

Or you don't delete enough and leave behind orphaned rows that confuse people / waste space / lead to weird behaviour :-)

Showing the rows and getting user confirmation is not the same thing (in my mind) as enforcing data correctness etc. We should do both; e.g. when deleting a domain the user should know they're also deleting 25 mailboxes and 15 aliases.

TrapoSAMA:
Would a useful first step be a **manual, read-only referential-integrity check** in `setup.php`, instead of immediately adding the foreign keys or deleting existing data?

The idea would be:

- Add a **Check database integrity** action to the authenticated Database section in `setup.php`.
- Run only `SELECT` queries and report the affected relation/table, row count, and enough identifiers to inspect the problem.
- Include guidance (and, where useful, inspection SQL) for the administrator to correct the data manually after making a backup.
- Handle legitimate/special cases explicitly, for example `domain_admins.domain = 'ALL'`, and treat externally written tables such as `quota`/`quota2` with care.
- Support MySQL/MariaDB, PostgreSQL, and SQLite via `table_by_key()`.
- Make the check repeatable and non-destructive; no automatic cleanup, schema changes, or foreign keys in this first PR.

I think this fits the concern raised above about not blindly deleting rows, while still giving administrators a practical way to discover inconsistencies left by partial migrations, interrupted operations, manual SQL changes, or external database writers.

[...]

PR #1091, opened 2026-07-23
Add a read-only database integrity check

TrapoSAMA:
## Summary

- add an authenticated **Check database integrity** action to `setup.php`
- detect orphaned rows for the six relationships discussed in #972
- keep the check read-only and repeatable, with row counts, limited sample identifiers, and manual remediation guidance
- exclude the special `domain_admins.domain = 'ALL'` assignment
- add SQLite-backed tests for detection, non-destructive behavior, and the special-domain case

## Why

This partially addresses #972 without deleting user data or adding foreign keys yet. It gives administrators a safe way to identify inconsistencies left by partial migrations, interrupted operations, manual SQL changes, or external database writers before deciding how to correct them.

The scope follows the approach discussed in https://github.com/postfixadmin/postfixadmin/issues/972#issuecomment-5052913570.

## Impact

- no schema-version change
- no automatic cleanup
- no foreign keys added
- supports configured table names through `table_by_key()`
- uses portable queries for MySQL/MariaDB, PostgreSQL, and SQLite

## Validation

- `DatabaseIntegrityCheckerTest.php`: 3 tests, 9 assertions
- PHPStan level 0 on the changed implementation, UI, and tests
- PHP lint and PHP-CS-Fixer checks on all changed PHP files

All targeted checks were run with the project PHP 8.4 runtime.

### DIFF
--- a/model/DatabaseIntegrityChecker.php
+++ b/model/DatabaseIntegrityChecker.php
@@ -1,0 +1,108 @@
+<?php
+
+class DatabaseIntegrityChecker
+{
+    /**
+     * @return array<int, array<string, mixed>>
+     */
+    public function check(int $sample_limit = 20): array
+    {
+        $sample_limit = max(1, min(100, $sample_limit));
+        $results = [];
+
+        foreach ($this->relations() as $relation) {
+            $child_table = table_by_key($relation['child_table']);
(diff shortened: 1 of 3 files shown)


DavidGoodwin:
I've not yet seen it moan about orphaned data etc

image

Issue #1064, opened 2026-07-01
Consolidate pagination onto one reusable Bootstrap 5 component (retire cNav_bar)


knofte:
### Tracking / refactor

PR #1063 introduces a standard Bootstrap 5 `.pagination` component for the View Log pager. PostfixAdmin ships Bootstrap 5.3 but currently hand-rolls pagination differently on every page. This issue tracks consolidating **all** pagination onto one reusable Bootstrap 5 component to cut duplicated markup.

### Current state — two paradigms

| Surface | Paradigm | Renderer |
| --- | --- | --- |
| `viewlog.php` | numeric pages (1, 2, 3 …) | Bootstrap 5 `.pagination` (done in #1063) |
| `list-virtual.php` (alias + mailbox) | alphabetical buckets (`aa-cz`, `de-pf` …) | the `cNav_bar` class + `<img>` arrow images |
| `list.php` (generic list) | alphabetical buckets via `getPagebrowser()` | template markup |

Note the numeric windowed helper `pagination_window()` is **numeric-only** and does not apply to the alphabetical page-browser — those page by name range, not page number. What is reusable across both is the Bootstrap `.pagination` **markup**.

### Proposed approach

Extract a shared Smarty partial `templates/_pagination.tpl` that renders a `.pagination` list from a normalized set of items (`{label, target, active, disabled}`). Each surface normalizes its own data (numeric pages from `pagination_window()`, or the alpha buckets) and `{include}`s the partial.

Payoff: the whole `cNav_bar` class (~120 lines of PHP building `<table>` + image-arrow HTML in `list-virtual.php`) can be deleted, and the `images/arrow-*.png` navigation images retired.

### Suggested breakdown (each its own PR)

1. Extract `templates/_pagination.tpl`; convert `viewlog.php` to use it (small; proves the partial).
2. Replace `cNav_bar` in `list-virtual.php` with the partial (the big win, most-used page). Careful: `cNav_bar` also carries the search-threading / `append_to_url` logic, so this is a real refactor, not find-replace.
3. Convert the generic `list.php` page-browser to the partial.

### Depends on

#1060 and #1063 landing first (this builds on the pattern introduced there).

knofte:
@cboltz I'd suggest I tackle this next, so we standardise this part. It's nothing major, just nice to have one way of things, instead each php file doing its own. :)

TrapoSAMA:
[...]

PR #1093, opened 2026-07-28
Replace list-virtual cNav_bar with shared pagination

TrapoSAMA:
## Summary

This is the second step of #1064, following #1090.

- replace `list-virtual.php`'s `cNav_bar` HTML renderer with normalized items for the shared `templates/_pagination.tpl`
- preserve the selected domain and nested search parameter in every page link
- use stable `#aliases` and `#mailboxes` anchors for the two independently rendered lists
- render the same accessible Bootstrap 5 pagination above and below each list
- remove the now-unused `arrow-l.png`, `arrow-r.png`, and `arrow-u.png` assets
- add focused tests for first, previous, range, next, query-string, and invalid-page-size behavior

The alphabetical ranges still come from the existing `create_page_browser()` logic. This change only normalizes those ranges for the shared renderer and removes the PHP-generated table/image markup.

## Validation

- focused pagination tests: 12 tests, 149 assertions
- PHP-CS-Fixer dry run: clean
- parallel PHP lint: clean
- Psalm: no errors
- Smarty compilation of `list-virtual.tpl`: clean
- authenticated XAMPP/browser checks with `page_size=20` and populated alias/mailbox lists:
  - first and second pages
  - active/disabled states and `aria-current`
  - stable section anchors
  - nested `search[_]` and domain propagation
  - desktop and mobile layouts
  - no console errors, broken images, or legacy arrow-image requests

The complete local suite was also exercised. The changed pagination tests passed; the remaining failures are the existing Windows/XAMPP environment-specific checks for Unix commands, DNS discovery, GD, and URL path escaping.


DavidGoodwin:
thanks @TrapoSAMA 

am I being old, or is there something not quite right with the pagination controls here? 

image

 * should the controls be at the top AND bottom of the list ?
 * can we stop the page jumping around when you go from one page to the next (using any of the bootstrap controls).
 * do we need to have some background colors applied or something so it's clearer which is the forward block ,and which is mailboxes etc? or can the pagination controls move within the :: Forwards block?

more like this ? 

list-virtual doctored a little

TrapoSAMA:
Thanks, you were right: the pagination controls were outside the section they controlled, so the anchor could land on the list while leaving the controls out of context.

I've moved both the top and bottom pagers inside the respective Forwards and Mailboxes cards. The existing card header and border now provide the visual grouping without adding new colors, and the existing #aliases / #mailboxes anchors return to the correct section when changing pages.

I validated both pagers with populated lists on desktop and mobile, including page navigation and horizontal overflow, with no browser console errors. The focused pagination tests pass on PHP 8.4 (11 tests, 37 assertions).

DavidGoodwin:
@TrapoSAMA can you fix the conflicts please?

TrapoSAMA:
@DavidGoodwin I rebased #1093 onto the current `master` and resolved the conflicts.

The overlap with #1127 was mechanical: `functions.inc.php`, `templates/list.tpl`, and `tests/PageBrowserPaginationTest.php` are already present upstream, so the rebased PR now keeps only the `list-virtual.php` migration, its three list-virtual templates, and the obsolete arrow-image deletions. The password-expiration changes already in `master` from #1109 were preserved.

Validation after the rebase:

- PHP 8.4 lint and parallel-lint passed for `public/list-virtual.php`.
- PHPStan level 0 and `git diff --check` passed.
- The regenerated textual production patch applies cleanly with GNU `patch -p1` against current `master`, without rejects, fuzz, offsets, or `.orig` files; the applied files match the rebased head exactly.

The existing pagination behavior and the prior green PR test matrix are unchanged; the local focused PHPUnit rerun was not completed because the validation host hit its tool-usage limit.

DavidGoodwin:
Thanks!

PR #1096, opened 2026-07-30
Distinguish mailbox and domain quota validation errors

TrapoSAMA:
## Summary

- distinguish the per-mailbox maximum quota error from insufficient remaining domain quota
- add the domain-quota error message to every language catalog
- add focused coverage for both quota failure causes

## Root cause

Mailbox quota validation used the same generic translation for both the mailbox maximum and the domain-wide quota limit, so administrators could not tell why a valid per-mailbox value was rejected.

## Scope

This pull request now contains only the quota validation change and its required language key. General Spanish translation corrections are tracked separately in #1100.

## Validation

- PHP 8.4 lint
- `git diff --check`
- PHP-CS-Fixer
- `MailboxHandlerTest`: 4 tests, 56 assertions
- GNU `patch -p1 --dry-run`
- authenticated XAMPP checks for mailbox quota and browser console errors

### DIFF
--- a/model/MailboxHandler.php
+++ b/model/MailboxHandler.php
@@ -11,6 +11,7 @@ class MailboxHandler extends PFAHandler
     protected string $id_field = 'username';
     protected ?string $domain_field = 'domain';
     protected array $searchfields = ['username'];
+    protected string $quota_error = 'pEdit_mailbox_quota_text_error';
 
     # init $this->struct, $this->db_table and $this->id_field
     protected function initStruct()
@@ -435,7 +436,7 @@ class MailboxHandler extends PFAHandler
     protected function _validate_quota($field, $val)
     {
         if (!$this->check_quota($val)) {
-            $this->errormsg[$field] = Config::lang('pEdit_mailbox_quota_text_error');
+            $this->errormsg[$field] = Config::lang($this->quota_error);
             return false;
         }
         return true;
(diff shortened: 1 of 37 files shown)


model/MailboxHandler.php
```diff
@@ -435,7 +436,11 @@ protected function _prefill_domain($field, $val)
     protected function _validate_quota($field, $val)
     {
         if (!$this->check_quota($val)) {
-            $this->errormsg[$field] = Config::lang('pEdit_mailbox_quota_text_error');
+            $message = Config::lang($this->quota_error);
+            if ($message === '') {
```
cboltz: Please add the new error message to all *.lang files instead of implementing a fallback to another message.

You can/should use the update-languages.sh script to update all the *.lang files.

TrapoSAMA: Thanks, you're right. I removed the runtime fallback and ran languages/language-update.sh --patch, so the new quota message is now present in every language file, with English # XXX placeholders where it is not translated. The script also propagated the six earlier alias-default keys it found missing. All language files pass PHP lint, MailboxHandlerTest passes with 4 tests and 56 assertions, PHP-CS-Fixer is clean, and all GitHub checks pass. I applied the corresponding quota-key correction to the 4.x companion PR #1097 as well.

TrapoSAMA:
Scope update: I rewrote the branch so this PR now contains only the quota validation change and its required language key. The unrelated Spanish translation corrections were moved to #1100.

PR #1097, opened 2026-07-30
Distinguish mailbox and domain quota validation errors on 4.x

TrapoSAMA:
## Summary

- distinguish the per-mailbox maximum quota error from insufficient remaining domain quota
- add the domain-quota error message to every language catalog
- add focused coverage for both quota failure causes

## Root cause

Mailbox quota validation used the same generic translation for both the mailbox maximum and the domain-wide quota limit, so administrators could not tell why a valid per-mailbox value was rejected.

## Scope

This pull request now contains only the quota validation change and its required language key. General Spanish translation corrections are tracked separately in #1101.

## Validation

- PHP 8.4 lint
- `git diff --check`
- PHP-CS-Fixer
- `MailboxHandlerTest`: 4 tests, 56 assertions
- GNU `patch -p1 --dry-run`
- authenticated XAMPP v4.0.5 checks for mailbox quota and browser console errors


TrapoSAMA:
Scope update: I rewrote the branch so this PR now contains only the quota validation change and its required language key. The unrelated Spanish translation corrections were moved to #1101.

PR #1105, opened 2026-08-08
Improve form focus and TOTP input semantics

TrapoSAMA:
## Summary

- use a softer, consistent primary-blue focus treatment for standard data-entry controls
- identify TOTP inputs as one-time codes and request numeric keyboards on supported devices
- keep TOTP digits visible while they are entered
- fix the TOTP setup input ID so its label points to the actual control

## Impact

The focus rule applies to `.form-control` and `.form-select`. Buttons, links,
checkboxes, and theme controls retain their existing focus styles. TOTP
validation and server-side behavior are unchanged.

## Validation

- `git diff --check`
- verified the focused diff contains only the shared stylesheet and three TOTP templates
- verified every `fTOTP_code` input uses `type="text"`, `inputmode="numeric"`, and `autocomplete="one-time-code"`

Authenticated browser validation was not run because the isolated browser
session did not have a PostfixAdmin login.

### DIFF
--- a/templates/totp.tpl
+++ b/templates/totp.tpl
@@ -25,7 +25,7 @@
             </div>
             <div class="mb-3 {if $pTOTP_code_text}is-invalid{/if}">
                 <label class="col-md-2" for="fTOTP_code">{$PALANG.pTOTP_code}:</label>
-                <div class="col-md-6 col-sm-8"><input id="fTOTP_code" class="form-control" type="password"
+                <div class="col-md-6 col-sm-8"><input id="fTOTP_code" class="form-control" type="text"
                                                       name="fTOTP_code" size="6" inputmode="numeric"
                                                       autocomplete="one-time-code"/>
                     <span class="text-warning">{$pTOTP_code_text}</span> <!-- error text -->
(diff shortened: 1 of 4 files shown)


templates/totp.tpl
```diff
@@ -25,8 +25,9 @@
             </div>
             <div class="mb-3 {if $pTOTP_code_text}is-invalid{/if}">
                 <label class="col-md-2" for="fTOTP_code">{$PALANG.pTOTP_code}:</label>
-                <div class="col-md-6 col-sm-8"><input id="TOTP_code" class="form-control" type="text" name="fTOTP_code"
-                                                      size="6"/>
+                <div class="col-md-6 col-sm-8"><input id="fTOTP_code" class="form-control" type="password"
```
DavidGoodwin: personally i'd find it annoying if the TOTP field was of type password; I'm not sure I can remember any I've used that star out the numbers.

DavidGoodwin (changes requested):

templates/totp.tpl
TrapoSAMA: Thanks, agreed. I changed all three TOTP inputs back to `type=text` so the digits remain visible, while retaining `inputmode=numeric`, `autocomplete=one-time-code`, and the corrected label/input IDs. TOTP validation is unchanged, and `git diff --check` passes.

DavidGoodwin (approved):

PR #1106, opened 2026-08-08
Improve TOTP code input semantics

TrapoSAMA:
## Summary

- identify TOTP inputs as one-time codes and request numeric keyboards on supported devices
- keep TOTP digits visible while they are entered
- fix the TOTP input IDs so their labels point to the actual controls

## Impact

This change is limited to the three TOTP templates. TOTP validation, session
handling, forms unrelated to TOTP, and application styling remain unchanged.

## Validation

- `git diff --check`
- verified the focused diff contains only the three TOTP templates
- verified every `fTOTP_code` input uses `type="text"`, `inputmode="numeric"`, and `autocomplete="one-time-code"`

Authenticated browser validation was not run because the isolated browser
session did not have a PostfixAdmin login.

### DIFF
--- a/templates/login-mfa.tpl
+++ b/templates/login-mfa.tpl
@@ -5,7 +5,7 @@
             <input class="flat" type="hidden" name="token" value="{$smarty.session.PFA_token|escape:"url"}"/>
             <div class="form-group {if $pTOPT_code_text}has-error{/if}">
                 <label class="col-md-4 col-sm-4 control-label" for="fTOTP_code">{$PALANG.pTOTP_code}:</label>
-                <div class="col-md-6 col-sm-8"><input id="fTOTP_code" class="form-control" type="password"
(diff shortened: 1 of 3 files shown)


DavidGoodwin:
I think I'm used to seeing TOTP code inputs as numerics - and not type=password. 

Personally I wouldn't want them starred out as I can't then correct a typo. I'm not quite sold on the 'someone might be viewing your screen' issue ... as they'd also have to know your username+password too. 

AWS don't use type=password on their input form - and that must be subject to more stringent security testing than postfixadmin?

aws-signin

TrapoSAMA:
Thanks for the AWS example. Agreed: all three TOTP inputs now remain visible with `type=text`, while retaining `inputmode=numeric`, `autocomplete=one-time-code`, and corrected label/input IDs. TOTP validation is unchanged, and `git diff --check` passes.

PR #1109, opened 2026-08-08
Fix password expiration handling and visibility

TrapoSAMA:
## Summary

- validate domain password-expiration policies as whole numbers from 0 through 36500 days
- make password expiration optional and disabled by default
- make 0 consistently mean that passwords in the domain do not expire
- align administrator, recovery, mailbox-user, and XML-RPC password changes
- show each mailbox password-expiration state in lists and CSV output
- add the new labels to every language file
- document how to enable the feature in `config.local.php`

## Behavior

- The default is `$CONF['password_expiration'] = 'NO';`.
- Administrators can opt in from `config.local.php`.
- Disabled or zero-day policies use the portable no-expiration sentinel.
- Positive policies calculate the timestamp when the password changes.
- Invalid historical values fail safe as no expiration and are logged.
- Timestamp-calculation failure raises an exception instead of silently changing policy.
- Domain policy changes remain non-retroactive.

## Validation

- focused PHP 8.4/SQLite suites: 33 tests, 155 assertions
- PHP lint passed for all 39 affected PHP and language files
- `git diff --check`
- GNU patch verification reproduced the current GitHub merge-ref tree exactly
- GitHub CI passed lint, PHP 8.2–8.5, MySQL, and PostgreSQL

The local PHP-CS-Fixer result was not used because the Windows checkout caused
whole-file line-ending differences.


functions.inc.php
```diff
@@ -297,6 +297,69 @@ function get_password_expiration_value(string $domain)
...
+ * A domain value of 0 means that passwords do not expire. The database still
+ * requires a timestamp, so use the latest portable DATETIME value as a
+ * sentinel. If the feature is disabled, preserve the historical 365-day
+ * fallback used for the non-null mailbox column.
+ */
+function get_mailbox_password_expiry(string $domain, ?int $now = null): string
+{
+    $days = 365;
+
+    if (Config::bool('password_expiration')) {
```
cboltz: If password_expiration is disabled, the function will return "365 days" instead of "never".

Better do the if condition the other way round:

```
    if (Config::bool('password_expiration')) {
        return PASSWORD_EXPIRATION_NEVER;
    }
```

functions.inc.php
```diff
@@ -297,6 +297,69 @@ function get_password_expiration_value(string $domain)
...
+        if ($days > PASSWORD_EXPIRATION_MAX_DAYS) {
+            error_log("password_expiry value for domain $domain exceeds the supported maximum; treating it as no expiration");
+            return PASSWORD_EXPIRATION_NEVER;
+        }
+    }
+
+    $base = $now ?? time();
+    $expiry = strtotime("+$days days", $base);
+    if ($expiry === false) {
+        return PASSWORD_EXPIRATION_NEVER;
```
cboltz: The documentation says:

> Returns a timestamp on success, false otherwise.

So if strtotime fails, users silently get a non-expiring password? Good for them, but an error message might be better ;-)

languages/en.lang
```diff
@@ -45,6 +45,8 @@ $PALANG['pFooter_logged_as'] = 'Logged in as %s';
 $PALANG['pLogin_welcome'] = 'Mail admins login here to administer your domain.';
 $PALANG['pLogin_username'] = 'Login (email)';
 $PALANG['password'] = 'Password';
+$PALANG['password_expiration_on'] = 'Expires on';
+$PALANG['password_expiration_never'] = 'Does not expire';
```
cboltz, line 49: Please add this to all *.lang files using the language-update.sh script.

model/MailboxHandler.php
```diff
@@ -312,14 +312,9 @@ protected function preSave(): bool
...
-            // provide some default value to keep MySQL etc happy.
-            $this->values['password_expiry'] = date('Y-m-d H:i', strtotime("+365 days"));
-            if (Config::bool('password_expiration')) {
-                $domain_dirty = $this->domain_from_id();
-                $domain = trim($domain_dirty, "`'"); // naive assumption it is ' escaping.
-                $password_expiration_value = (int)get_password_expiration_value($domain);
-                $this->values['password_expiry'] = date('Y-m-d H:i', strtotime("+$password_expiration_value day"));
-            }
+            $domain_dirty = $this->domain_from_id();
+            $domain = trim($domain_dirty, "`'"); // naive assumption it is ' escaping.
```
cboltz: I know that this was there before, but it's still silly nevertheless.

If `$this->domain_from_id()` ever returns something that is in quotes, it is broken.

Please remove the trim(...) - and in the unlikely case that something breaks, fix it where it's broken.

public/list.php
```diff
@@ -131,6 +131,12 @@
     fputcsv($out, $header);
 
     # print items as csv
+    $password_expiration_policies = array();
+    $password_expiration_never_label = Config::lang('password_expiration_never');
+    if ($password_expiration_never_label === '') {
+        $password_expiration_never_label = Config::lang('pOverview_unlimited');
+    }
+
```
cboltz: IMHO list.php should never contain field-specific logic. Please move it to the *Handler class that needs it.

templates/list-virtual_mailbox.tpl
```diff
@@ -24,6 +24,9 @@
         {if $CONF.quota===YES}
             <th>{$PALANG.pOverview_mailbox_quota}</th>{/if}
         <th>{$PALANG.last_modified}</th>
+        {if $password_expiration}
+            <th class="text-center text-nowrap">{$PALANG.password}</th>
+        {/if}
```
cboltz, line 29: Unrelated to this MR, but I'll still mention it:

Ideally listing mailboxes should be done using list.tpl. Obviously we'll need to add some field-specific handling there - that's expected and fine.

Just in case you are bored, this would be a nice follow-up MR (wait until this one is handled to avoid merge conflicts).

tests/DomainHandlerTest.php
```diff
@@ -2,6 +2,61 @@
...
+                );
+                $this->assertSame((int)$value, (int)$stored['password_expiry']);
+            }
+        } finally {
+            db_delete('domain_admins', 'domain', $domain);
+            db_delete('domain', 'domain', $domain);
+        }
+    }
+
+    public static function passwordExpiryValuesProvider(): array
```
cboltz, line 45: Please also add a testcase for cases when password expiry is disabled in the config.

Speaking about this: NIST and others recommend to get rid of expiring passwords, see for example [this article](https://nextperimeter.com/insights/why-forced-password-changes-every-90-days-are-no-longer-best-practice/)

Also, the latest shadow utils (4.20.0) removed support for password expiration completely.

Therefore I wonder if we should follow this and _at least_ change `$CONF['password_expiration'] = 'YES';` to default to `NO`. (I'm not sure if I should propose to follow the shadow utils way... ;-)

tests/PasswordExpirationTest.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->assertSame(PASSWORD_EXPIRATION_NEVER, get_mailbox_password_expiry($this->domain, 0));
+    }
+
+    public function testDisabledFeatureUsesSameFallbackForEveryChangePath(): void
+    {
+        Config::write('password_expiration', 'NO');
+        $base = strtotime('2026-01-01 00:00');
+
+        $this->assertSame(
+            date('Y-m-d H:i', strtotime('+365 days', $base)),
```
cboltz: Ah, here's the test I asked for.

However, the expectation is wrong - if expiry is disabled, the expected value should be "never".

public/list-virtual.php
```diff
@@ -230,7 +231,8 @@
 }
 
 if ($password_expiration) {
-    $sql_select .= ", $table_mailbox.password_expiry as password_expiration ";
+    $sql_select .= ", $table_mailbox.password_expiry as password_expiration, $table_domain.password_expiry as domain_password_expiration ";
+    $sql_join .= " LEFT JOIN $table_domain ON $table_mailbox.domain=$table_domain.domain ";
```
cboltz, line 235: Another note for a separate MR whenever you are bored:

Ideally the logic for listing mailboxes, creating the SQL etc. should live in MailboxHandler initStruct().

Needless to say that this isn't as easy as it might (or might not) look, which is probably the reason why nobody did it yet ;-)

functions.inc.php
TrapoSAMA: Thanks, agreed. When password expiration is disabled, `get_mailbox_password_expiry()` now returns `PASSWORD_EXPIRATION_NEVER` instead of the historical 365-day fallback. The focused test now covers this behavior.

functions.inc.php
TrapoSAMA: Agreed. A failed expiration calculation now raises a `RuntimeException` instead of silently converting the password to non-expiring.

languages/en.lang
TrapoSAMA: Thanks. Both new password-expiration keys are now present in every language catalog. Existing English and Spanish translations are retained, while untranslated catalogs receive the English text marked with `# XXX`. PHP lint passes for all affected catalogs.

model/MailboxHandler.php
TrapoSAMA: Agreed. The unnecessary `trim()` has been removed, and the value returned by `domain_from_id()` is now used directly.

public/list.php
TrapoSAMA: Agreed. I removed the password-expiration-specific logic from `public/list.php` and moved the display normalization into `MailboxHandler::read_from_db_postprocess()`, so both lists and CSV exports use the shared handler result. A focused regression test covers no-expiration formatting.

templates/list-virtual_mailbox.tpl
TrapoSAMA: Good point, thanks. I agree that moving the mailbox overview to the generic list template would be useful, but I am keeping that broader refactor in a separate follow-up PR so this correction remains focused and avoids merge conflicts.

tests/DomainHandlerTest.php
TrapoSAMA: Thanks, agreed. Password expiration now defaults to `NO`, and the disabled-feature coverage expects the no-expiration sentinel. Administrators who need the feature can opt in from `config.local.php`.

tests/PasswordExpirationTest.php
TrapoSAMA: Corrected. When password expiration is disabled, the test now expects `PASSWORD_EXPIRATION_NEVER`. The Login test also explicitly enables the feature for its separate 30-day scenario and restores the original configuration afterward.

public/list-virtual.php
TrapoSAMA: Agreed. Moving the remaining custom mailbox-list query into `MailboxHandler::initStruct()` is a useful follow-up, but I am keeping it outside this PR to avoid mixing the behavioral fix with a broader listing refactor.

cboltz:
Thanks for all the updates! This MR looks good now :-)

@DavidGoodwin: if you agree, feel free to merge.

PR #1110, opened 2026-08-08
Fix password expiration handling and visibility on 4.x

TrapoSAMA:
## Summary

- port the password-expiration correction to `postfixadmin_4.0`
- make password expiration optional and disabled by default
- make 0 consistently mean that domain passwords do not expire
- validate whole-day policies from 0 through 36500
- align password-change paths and preserve the existing Login fallback
- show mailbox expiration states in lists and CSV output
- add the new labels to every 4.x language file
- document how to enable the feature in `config.local.php`

## Compatibility

- targets `postfixadmin_4.0`
- keeps PHP 7.4-compatible syntax and Bootstrap 3 conventions
- leaves the feature disabled unless an administrator explicitly enables it

## Validation

- password-expiration, mailbox, and Login suites on PHP 8.4/SQLite: 17 tests, 98 assertions
- PHP lint passed for all 38 affected PHP and language files
- `git diff --check`
- GNU patch verification reproduced the current GitHub merge-ref tree exactly
- branch CI passed the PHP 7.4 and 8.x matrix

The historical 4.x `DomainHandlerTest` data provider is non-static and PHPUnit
11 rejects that test before execution. Branch CI remains the authoritative
PHP 7.4/8.x validation.


TrapoSAMA:
The same review corrections have been ported to 4.x: password expiration is disabled by default, administrators can opt in from `config.local.php`, disabled passwords do not expire, display normalization lives in the handler, and all 4.x language catalogs include the new labels. The Login test now explicitly enables expiration for its 30-day scenario and restores the original configuration. Branch CI passed on PHP 7.4 and 8.x.

PR #1111, opened 2026-08-08
Document optional Dovecot mail deletion logging

TrapoSAMA:
## Summary

- include optional `mail_log_events` and `mail_log_fields` settings in the Dovecot 2.4 example
- keep detailed logging and the required `mail_log`/`notify` plugins commented out by default
- explain how administrators can explicitly enable the complete set
- link to the official Dovecot documentation
- document the prospective nature and privacy implications of sender and subject fields

## Why

Detailed deletion logging can help investigate future reports of missing mail,
but sender and subject metadata may be sensitive. The example documents the
capability without enabling it by default.

## Validation

- `git diff --check`
- verified all related settings remain commented out in the example
- verified the events and fields against the official Dovecot 2.4.1 documentation

`doveconf -n` was not run because Dovecot is unavailable on the Windows host.

### DIFF
--- a/DOCUMENTS/examples/dovecot-2.4-local.conf
+++ b/DOCUMENTS/examples/dovecot-2.4-local.conf
@@ -181,6 +181,11 @@ mail_plugins {
   fts_flatcurve = yes
 }
 
+# Record message and mailbox deletion activity for operational troubleshooting.
+# The mail_log and notify plugins are enabled for IMAP below.
+mail_log_events = delete undelete expunge mailbox_delete
+mail_log_fields = uid box msgid size from subject
+
 acl_driver = vfile
 
 acl_sharing_map {
--- a/DOCUMENTS/examples/dovecot-2.4-local.conf
+++ b/DOCUMENTS/examples/dovecot-2.4-local.conf
@@ -181,10 +181,10 @@ mail_plugins {
   fts_flatcurve = yes
 }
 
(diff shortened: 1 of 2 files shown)


DOCUMENTS/examples/dovecot-2.4-local.conf
```diff
@@ -181,6 +181,11 @@ mail_plugins {
   fts_flatcurve = yes
 }
 
+# Record message and mailbox deletion activity for operational troubleshooting.
+# The mail_log and notify plugins are enabled for IMAP below.
+mail_log_events = delete undelete expunge mailbox_delete
+mail_log_fields = uid box msgid size from subject
```
cboltz: I'm a fan of privacy by default, and especially logging the subject doesn't look like the best idea to me. Still, I can imagine cases where it makes sense - but IMHO it shouldn't be enabled by default.

Would it make sense to comment out these lines in the default config so that admins have to explicitely enable them if they wish this detailed (and not privacy-friendly) logging?

TrapoSAMA: Thanks, agreed. The detailed `mail_log_events`, `mail_log_fields`, `mail_log`, and `notify` settings are now commented out by default. The documentation presents deletion logging as an explicit opt-in and asks administrators to assess whether logging sender and subject metadata is appropriate for their privacy and retention policy.

DavidGoodwin (approved):

PR #1114, opened 2026-08-11
Harden language catalog repair and syntax validation

### LINKED ISSUE #1112: language-update.sh adds duplicated strings
The rewritten language-update.sh proposes to add two seemingly missing strings to several *.lang files.

After actually updating *.lang with --patch, it instantly reports these strings as duplicates.

```
# ./language-update.sh bg.lang
### bg.lang: 2 missing, 0 obsolete ###
$PALANG['password_expiration_desc'] = 'Number of days before the password expires'; # XXX   after 'password_expiration'
$PALANG['To_Mailbox']                       = 'Mailbox'; # XXX   after 'password_expiration'
# ./language-update.sh bg.lang --patch
*** bg.lang: added 2 missing string(s) ***
      2 password_expiration_desc
      2 To_Mailbox
*** duplicated string in bg.lang, see above for details ***
```

While I only show bg.lang here, more languages are affected. In total, this bug happens for 23 *.lang files. OTOH, 12 *.lang files don't trigger this bug.

### LINKED ISSUE #1113: Make language-update.sh run part of CI
When MRs add new UI strings, they often change only one or two *.lang files instead of updating all of them. (As demonstrated by some recent PRs - I just fixed this in master.)

Ideally the CI should catch and report that.

This needs two changes:
- language-update.sh: set a non-zero exitcode if translations are missing or obsolete in any *.lang file (currently it always exits with $?=0)
- CI: run language-update.sh and fail if the exitcode is not 0

TrapoSAMA:
## Summary

- validate each language catalog before comparing translation keys
- abort the catalog scan immediately when PHP syntax is invalid
- repair the existing missing semicolons in 24 catalogs as a one-off correction
- return a non-zero status for missing or obsolete translations
- run both the catalog comparison and explicit `php -l` checks in CI
- add focused regression coverage

## Root cause

The tokenizer-based parser accepted token streams from syntactically invalid catalogs. Missing semicolons caused adjacent `$PALANG` assignments to be treated as one statement, so `--patch` could insert duplicate keys.

## Behavior

`languages/language-update.sh` does not attempt automatic repair. It validates the complete PHP source before comparing translation keys and aborts the scan on the first syntax error.

Read-only catalog checks return status 1 when any catalog has missing or obsolete keys. `--patch` can return 0 after adding every missing key, but still returns 1 when obsolete keys remain. The GitHub Actions lint job runs the complete catalog check and explicitly executes `php -l` for every language file.

## Validation

- PHP 8.4 lint passed for all 36 language catalogs and the affected PHP test file
- Bash syntax check passed
- the complete 36-catalog check returned 0
- invalid syntax returned 1 and prevented later catalogs from being processed
- `git diff --check` passed
- PostfixAdmin 4.x was inspected separately and does not contain this syntax defect
- GitHub CI run 31548069597 passed lint, PHP 8.2-8.5, MySQL, and PostgreSQL

The full PHPUnit suite was not run locally because dependencies are unavailable in this checkout. Regression tests are included for CI.

Fixes #1112.
Fixes #1113.


DavidGoodwin:
I think we'd be better off doing a one off fix on those files with invalid syntax, and just aborting/breaking within update-languages.sh if there is invalid syntax in the future (as in, stop trying to be clever fixing up invalid syntax).

TrapoSAMA:
Thanks, agreed. I simplified languages/language-update.sh so it only validates catalog PHP syntax and aborts the scan on the first invalid file; the 24 affected catalogs remain a one-off correction. The optional repair logic is now isolated in ADDITIONS/fix-language-syntax.php and is never called by the main script or CI. The updated GitHub CI run passed lint, PHP 8.2 through 8.5, MySQL, and PostgreSQL.

ADDITIONS/fix-language-syntax.php
```diff
@@ -0,0 +1,178 @@
...
+    return [$source, [
+        'missing assignment semicolon' => $simple,
+        'missing heredoc semicolon' => $heredoc,
+    ]];
+}
+
+function next_backup_filename(string $file): string
+{
+    $backup = "$file.bak";
+    for ($suffix = 2; file_exists($backup); $suffix++) {
```
cboltz, line 86: file_exists comes with some funny details:
- it returns false on dead symlinks (so you can have symlinks attack even without any race involved)
- it caches the result

This is probably not very problematic for this script, but maybe you should at least add a comment  (in the file header) like "don't run this script with files in directories that are writeable by others (like /tmp/)"

ADDITIONS/fix-language-syntax.php
```diff
@@ -0,0 +1,178 @@
...
+    [$fixedSource, $fixCounts] = apply_safe_fixes($source);
+    $fixedTotal = array_sum($fixCounts);
+    $remainingError = php_syntax_error($fixedSource);
+    if ($fixedTotal === 0 || $remainingError !== null) {
+        fwrite(STDERR, "*** $file: no safe complete repair; file left unchanged ***\n");
+        return false;
+    }
+
+    $backup = next_backup_filename($file);
+    if (!copy($file, $backup)) {
```
cboltz, line 120: This could be affected by a race condition (or even without a race condition on dead symlinks).

Again, <copy&paste>this is probably not very problematic for this script, but maybe you should at least add a comment like "don't run this script with files in directories that are writeable by others (like /tmp/)"</copy&paste>

ADDITIONS/fix-language-syntax.php
```diff
@@ -0,0 +1,178 @@
...
+        fwrite(STDERR, "*** $file: unable to create backup $backup; file left unchanged ***\n");
+        return false;
+    }
+
+    $written = file_put_contents($file, $fixedSource);
+    if ($written !== strlen($fixedSource)) {
+        if (!copy($backup, $file)) {
+            fwrite(STDERR, "*** $file: write and backup restoration failed ***\n");
```
cboltz, line 128: Mentioning $backup as part of this error message would be helpful - assuming $backup contains the previous state, users can manually restore it.

cboltz:
> PostfixAdmin 4.x was inspected separately and does not contain the syntax defect

I introduced these syntax errors when I updated the *.lang files two days ago. The most obvious (and most boring) issue was that some strings were missing in most language files.

There were also some texts that were changed, and not yet translated in most languages. Instead of just adding a comment, I replaced the not-yet-translated texts with the new text using some script magic, and accidentally killed the semicolon. Sorry for that!

(I'm not sure if my translation updates + fixes from this PR are also needed in the 4.0 branch - feel free to review and backport them.)

ADDITIONS/fix-language-syntax.php
DavidGoodwin: the cache only lasts as long as the php process, so probably not a problem here.

I just don't see any need to have the 'fix-language-syntax.php' script in git, that's all. 

What we do probably need is to just run 'php -l' on each language file (or include languages/ in what psalm  or parallel-lint checks).

TrapoSAMA:
Thanks. I removed the automatic repair helper from the repository and added an explicit `php -l` check for every language catalog in CI. Could you confirm whether you prefer the one-off semicolon corrections in the affected `.lang` files to remain in this PR, or whether those files will be corrected internally? Separately, I will review Christian's observation about the 4.0 branch and submit a dedicated 4.x PR if a branch-specific correction is confirmed.

PR #1124, opened 2026-08-26
Document optional strict MX policy service

TrapoSAMA:
## Summary

Add a short documentation pointer from `ADDITIONS/README.md` to the standalone [postfix-strict-mx-policy](https://github.com/TrapoSAMA/postfix-strict-mx-policy) project.

The policy service implementation, tests, CI, installation instructions, and license now live exclusively in that independent repository. This PR contains no executable code and makes no runtime changes to PostfixAdmin.

The text also calls out that requiring an explicit MX record is intentionally stricter than standard SMTP implicit-MX delivery, so administrators can make an informed choice before enabling it.

## Validation

- `git diff --check`: passed.
- Standalone project CI: passed on Python 3.8 and 3.13.

No PHP runtime, database engine, schema, authorization flow, or PostfixAdmin configuration is affected.

### DIFF
--- a/ADDITIONS/README.md
+++ b/ADDITIONS/README.md
@@ -109,3 +109,10 @@ and recovery guidance.
 ## Cyrus Quota Usage
 
 See https://github.com/o-m-d/cyrus-quotausage-to-pfa
+
+## Optional strict MX policy service
+
+[postfix-strict-mx-policy](https://github.com/TrapoSAMA/postfix-strict-mx-policy)
+is a standalone Postfix policy service that rejects recipient domains without
+an explicit MX record. This is intentionally stricter than standard SMTP
+implicit-MX delivery; see the project documentation before enabling it.


DavidGoodwin:
I like the idea, but I think it's outside the scope of postfixadmin - the script doesn't use postfixadmin's database, it's not integrated into the UI etc - the only link is 'postfix'. 

So, while it may be of use to other postfix users, i think it might be better off in a standalone project / repo.

TrapoSAMA:
Thanks, that makes sense. I initially placed it under `ADDITIONS` because it is an optional Postfix companion service, but I agree that it has no direct PostfixAdmin integration.

I have moved it to a standalone repository so it can be maintained and used independently by other Postfix installations:

https://github.com/TrapoSAMA/postfix-strict-mx-policy

The standalone CI passes on Python 3.8 and 3.13.

Would you be open to keeping only a short documentation pointer in `ADDITIONS/README.md` for administrators who may find the optional service useful? If so, I can reduce this PR to that single documentation change. Otherwise, I will close the PR as-is.

DavidGoodwin:
Yes, a pointer to the project is fine.

TrapoSAMA:
Thanks. I have reduced the PR to the approved documentation pointer in `ADDITIONS/README.md`.

The service implementation and tests now live exclusively in the standalone repository. The PostfixAdmin PR contains no executable code or runtime changes. `git diff --check` passes, lint is already green, and the remaining standard CI jobs are running.

Issue #1128, opened 2026-08-29
Harden authentication, recovery, SMTP, and integration boundaries


TrapoSAMA:
A focused review of current `master` found several independent security-sensitive and reliability gaps. This issue intentionally omits reproduction details; fixes should be delivered as small, separately reviewable pull requests.

Proposed work:

- Require a fully authenticated role for TOTP-management pages and verify the current password before changing or removing a secret.
- Build password-recovery links from a validated canonical URL, keep observable responses uniform, and add persistent throttling suitable for concurrent requests.
- Apply the same MFA policy to the optional XML-RPC interface, or reject password-only XML-RPC sessions for accounts requiring a second factor.
- Make the SMTP client abort on every unexpected reply and require successful TLS negotiation before authentication.
- Make the documented Roundcube TOTP synchronization update exactly the selected account and deserialize preferences without enabling object construction.
- Escape audit-log values for their HTML and JavaScript contexts and remove unsafe inline interpolation.

Each pull request should include focused regression coverage and preserve the existing configuration defaults. The current scope is `master`. The maintained 4.x branch should be reviewed separately and ported only where the same flow exists.

PR #1131, opened 2026-08-29
Enforce TOTP policy for XML-RPC login

TrapoSAMA:
## Summary

- reject password-only XML-RPC login when the mailbox requires TOTP
- preserve XML-RPC login behavior for mailboxes without TOTP
- centralize the password-only eligibility check in `TotpPf`
- add focused regression coverage for valid, TOTP-protected, and invalid-password cases

The XML-RPC protocol has no second-factor exchange, so protected accounts fail closed.

## Validation

- focused PHPUnit regression: 1 test, 3 assertions
- PHP-CS-Fixer on all changed files
- PHP lint on the XML-RPC endpoint
- `git diff --check`

Part of #1128.

### DIFF
--- a/config.inc.php
+++ b/config.inc.php
@@ -790,6 +790,8 @@ $CONF['theme_custom_css'] = '';
 // XMLRPC Interface.
 // This should be only of use if you wish to use e.g the
 // Postfixadmin-Squirrelmail package
+// XML-RPC has no TOTP challenge flow. Mailbox accounts with a TOTP
+// secret cannot authenticate through this interface.
 //  change to boolean true to enable xmlrpc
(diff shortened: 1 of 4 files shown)


DavidGoodwin:
does there need to be some documentation stating that xmlrpc + totp don't get along ?

DavidGoodwin (approved):

TrapoSAMA:
Thanks, good point. XML-RPC only provides a username/password login flow and has no way to exchange a TOTP code. Accounts with TOTP therefore intentionally fail authentication through this interface after this fix.

I documented that limitation next to `xmlrpc_enabled` in 9e08ddda.

Validation: PHP lint for `config.inc.php` and `git diff --check` pass.

PR #1132, opened 2026-08-29
Enforce SMTP protocol state

TrapoSAMA:
Reject SMTP delivery when a required server response, connection write, or STARTTLS negotiation fails.

This prevents the caller from treating an incomplete delivery as successful. It also handles partial socket writes.

Tests: focused PHPUnit tests, PHP lint, PHP-CS-Fixer, and diff checks.

Refs #1128

### DIFF
--- a/functions.inc.php
+++ b/functions.inc.php
@@ -1371,42 +1371,49 @@ function smtp_mail(string $to, string $from, string $subject_or_data, ?string $b
         return false;
     }
 
-    smtp_expect($socket, 220);
+    try {
+        smtp_require_response($socket, 220);
 
-    fputs($socket, "EHLO $helo\r\n");
-    smtp_expect($socket, 250);
+        smtp_write($socket, "EHLO $helo\r\n");
+        smtp_require_response($socket, 250);
 
-    if ($type === 'starttls') {
-        fputs($socket, "STARTTLS\r\n");
-        smtp_expect($socket, 220);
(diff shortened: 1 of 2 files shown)


functions.inc.php
```diff
@@ -1428,6 +1435,31 @@ function smtp_expect($socket, $code)
...
+{
+    if (!smtp_expect($socket, $code)) {
+        throw new RuntimeException("Unexpected SMTP response; expected $code");
+    }
+}
+
+function smtp_write($socket, string $data): void
+{
+    $remaining = $data;
+    while ($remaining !== '') {
```
DavidGoodwin, line 1448: what does this gain us over just having `fputs($socket, "whatever)` ?

is there some edge case where we only write part of the string to the socket AND there's no error raised?

TrapoSAMA: Thanks for raising this. Yes: `fputs()` is an alias of `fwrite()`, and PHP documents that a network-stream write may finish after writing only part of the supplied string. In that case it returns the number of bytes written rather than necessarily raising an error.

The previous calls ignored that return value, so a truncated command or, more importantly, a truncated DATA body could be treated as sent. `smtp_write()` completes partial writes and converts `false`, a zero-byte write, or an exception into the existing controlled failure path. The edge case is uncommon for short SMTP commands, but it is valid for network streams and arbitrary-sized message bodies, so I think retaining the helper is warranted.

Reference: https://www.php.net/manual/en/function.fwrite.php

PR #1144, opened 2026-09-02
Add OIDC authentication support

doneisgood:
## Summary

Adds native OpenID Connect (OIDC) authentication as an alternative to local username/password auth. Users can log in via an external OIDC provider (e.g., Keycloak, Auth0, Okta) with auto-provisioning of admin accounts.

Closes: (no existing issue — new feature)

## Features

- **OIDC login** — users click "Login with SSO" and authenticate via external provider
- **Auto-provisioning** — new admin accounts created automatically on first OIDC login
- **JWT validation** — uses `firebase/php-jwt` with JWKS key fetching and X.509 chain validation
- **CSRF protection** — state parameter prevents cross-site request forgery
- **Nonce verification** — prevents token replay attacks
- **Configurable button text** — customize the SSO button label (e.g., "FrogPond.Cloud Keycloak")
- **Graceful fallback** — local password auth remains available as default

## How it works

1. User clicks "Login with SSO" on login page
2. PostfixAdmin discovers OIDC provider configuration (`.well-known/openid-configuration`)
3. User is redirected to provider's authorization endpoint
4. After authentication, provider redirects back to `oidc_callback.php`
5. Callback validates state, exchanges code for tokens
6. ID token is validated using `firebase/php-jwt` with JWKS keys from provider
7. If auto-provisioning is enabled and user doesn't exist, admin account is created
8. Session is initialized and user is redirected to main page

## Files changed

- `model/OIDC.php` — OIDC class (discovery, authorize, callback, JWT validation)
- `public/oidc_login.php` — initiates redirect to OIDC provider
- `public/oidc_callback.php` — handles callback, validates tokens, auto-provisions users
- `public/login.php` — added OIDC login button assignment
- `templates/login.tpl` — added OIDC button to login form
- `config.inc.php` — added `auth_provider`, `oidc_auto_provision`, and `oidc` config options

## Configuration

In `config.local.php`:

```php
$CONF['auth_provider'] = 'oidc';  // 'local' or 'oidc'
$CONF['oidc_auto_provision'] = true;  // auto-create admin users on first login
$CONF['oidc'] = array(
    'client_id'     => 'your-client-id',
    'client_secret' => 'your-client-secret',
    'issuer_url'    => 'https://keycloak.example.com/realms/your-realm',
    'redirect_uri'  => 'https://mailadmin.example.com/oidc_callback.php',
    'scopes'        => 'openid email profile',
    'login_button_text' => 'Login with SSO',
);
```

## Test plan

- [ ] Verify OIDC login redirect works (provider discovery + authorization URL)
- [ ] Verify callback validates state and nonce correctly
- [ ] Verify JWT validation with real provider tokens
- [ ] Verify auto-provisioning creates admin account on first login
- [ ] Verify existing admin accounts can still log in via OIDC
- [ ] Verify session initialization and redirect to main page
- [ ] Verify local password auth still works when `auth_provider = 'local'`
- [ ] Verify `config.local.php` is preserved during any future upgrades

## Local QA Results

| Check | Result |
| --- | --- |
| PHP lint | Pass |
| OIDC redirect to Keycloak | Verified |
| OIDC callback + session init | Verified (manual login) |
| Auto-provisioning | Verified (new user created on first login) |
| SMTP AUTH (Dovecot SQL) | Verified (password change → IMAP/SMTP works) |
| Local password auth fallback | Verified (mailadmin@ still logs in with password) |

## Use Case

This was developed for a homelab/family mail server (FrogPond.Cloud) where family members need to manage their own mailboxes without sharing the main admin account. OIDC via Keycloak SSO provides secure, individual access.

## Test Environment

- **OIDC Provider:** Keycloak (LXC 103, FrogPond.Cloud realm)
- **Database:** PostgreSQL with pgBouncer
- **SMTP:** Postfix with Dovecot SASL
- **IMAP:** Dovecot

## Development

Code was written by AI (Claude) and tested by human. All changes were deployed to a live system and verified with real user logins.


doneisgood:
Note: "Auto-provisioning — new admin accounts created automatically on first OIDC login" failed to mention that this does not call AdminHandler but inserts o the databsae (which I don't like). If i can get that to work, I'll amend the PR.

model/OIDC.php
```diff
@@ -0,0 +1,207 @@
+<?php
+
+/**
+ * Class OIDC
+ *
+ * Handles OpenID Connect authentication flow for PostfixAdmin.
+ * Uses firebase/php-jwt for JWT validation.
```
DavidGoodwin, line 7: firebase/php-jwt should be added to composer.json?

public/login.php
```diff
@@ -47,23 +47,28 @@
 
 if ($_SERVER['REQUEST_METHOD'] == "POST") {
 
-    CsrfToken::assertValid(safepost('CSRF_Token'));
+    if (!isset($_SESSION['PFA_token'])) {
```
DavidGoodwin: this change shouldn't be being undone. (we **should** be using the CsrfToken class here)

I suspect your changes are based on an older version of PostfixAdmin (or perhaps PostfixAdmin 4.0?)

public/login.php
```diff
@@ -47,23 +47,28 @@
...
+        pfa_handle_invalid_token();
+    }
+
 
     $lang = safepost('lang');
     $fUsername = trim(safepost('fUsername'));
     $fPassword = safepost('fPassword');
 
-    if ($lang != Languages::check_language(false)) { # only set cookie if language selection was changed
+    if ($lang != check_language(false)) { # only set cookie if language selection was changed
```
DavidGoodwin: ditto the above - we **should** be using Languages:: here.

public/login.php
```diff
@@ -93,23 +98,34 @@
...
-} elseif (isset($_SESSION['sessid'])) {
-    // Preserve anonymous sessions so multiple login forms keep their CSRF tokens.
-    // An existing authenticated session should still be cleared before login.
+} else {
     session_unset();
     session_destroy();
     session_start();
 }
 
+$_SESSION['PFA_token'] = md5(uniqid("pfa" . rand(), true));
```
DavidGoodwin: drop this - stale/old - see above.

DOCUMENTS/OIDC-Feature.md
```diff
@@ -0,0 +1,89 @@
+## Summary
```
DavidGoodwin, line 1: doc should possibly be at DOCUMENTS/OIDC-Feature.md or something similar.

model/OIDC.php
```diff
@@ -0,0 +1,207 @@
...
+        $url = $this->discovery['authorization_endpoint'] . '?' . http_build_query($params);
+        header('Location: ' . $url);
+        exit;
+    }
+
+    public function handleCallback(string $code, string $state): array|false
+    {
+        // Validate state
+        if (!isset($_SESSION['oidc_state'])) {
+            error_log('OIDC: No state in session');
```
DavidGoodwin: while I love logging, I suspect you have too many error_log() calls in this PR/code?

I'd be OK with any error_log calls that _could_ be used by something like fail2ban as a basis for detecting/blocking malicious requests (and therefore need blocking). 

I'm way of the data that is being logged - e.g. with the tokenResponse below - it's not an error - it's really you doing this for debug/development purposes ....

DavidGoodwin:
Thanks for the PR - I hope the comments above are of some use. I've not tried running the code yet. I can see how it'd be useful (assuming it can also be used for google/azure type logins).

DOCUMENTS/OIDC-Feature.md
doneisgood: Thank you!  Addressed

public/login.php
doneisgood: OK, rebased to the master and made sure i had the least amount of changes.

model/OIDC.php
doneisgood: Let over from the debugging step.

doneisgood:
Reduced debug logging per review. Removed 11 error_log calls that were development/debug only. Kept 4 security-relevant logs (CSRF, session fixation, JWT failure, nonce mismatch) that can be used by fail2ban to detect malicious activity.

model/OIDC.php
doneisgood: Found more excessive logging.

TrapoSAMA:
Thanks for contributing OIDC support. The generic discovery and authorization-code flow is structurally suitable for providers such as Microsoft Entra ID / Microsoft 365, Google, Okta, Auth0, and Keycloak, but I think the intended scope and some provider-compatibility and security details should be clarified before merging.

As implemented, `auth_provider = oidc` enables one global OIDC provider while local password authentication remains available to all administrators. This can work well when multiple managed domains share one Microsoft Entra tenant or another common IdP. It can also support a hybrid deployment where administrators from that IdP use OIDC and everyone else continues using local credentials.

However, this is not per-domain or multitenant OIDC. It does not support different issuers or client configurations for different domains or groups of domains. That would require provider selection and explicit mapping of OIDC identities or groups to PostfixAdmin domains and permissions.

Auto-provisioning currently creates an administrator without assigning any domains, so the new account cannot meaningfully administer the intended domains until a global administrator configures it manually. It also bypasses `AdminHandler`; this should preferably use the established handler and default to disabled unless there is an explicit authorization or domain-mapping policy.

A few security and correctness issues appear to need attention:

- Validate the ID token's `iss` against the configured issuer and `aud` against the client ID. Signature and expiration validation alone are insufficient.
- Require the nonce claim to be present and match the stored nonce. The current condition accepts a token with no nonce.
- Require an appropriate verified-email policy and consider binding accounts using the stable `issuer + sub` identity instead of email alone.
- The inactive-account check is ineffective because `AdminHandler::view()` does not filter by `active`; an explicitly disabled administrator may still log in through OIDC.
- PostfixAdmin currently uses `SameSite=Strict` for its session cookie. Cross-site callbacks from providers such as Microsoft, Google, Okta, or Auth0 may therefore lose the session containing `oidc_state`. The same-site Keycloak test may not expose this problem.
- Validate that the discovery document's `issuer` exactly matches the configured issuer and require appropriate HTTPS endpoints before sending credentials or tokens.
- When UserInfo is used, require its `sub` to match the ID token's `sub` before merging claims; UserInfo should not be able to silently replace authenticated identity claims such as `email`.

Delegating MFA to the external provider is reasonable and normally desirable. OIDC can intentionally replace PostfixAdmin's local TOTP, but this behavior should be documented so administrators understand that MFA enforcement becomes the IdP's responsibility. Provider-specific `acr` or `amr` enforcement can remain an optional policy rather than a universal requirement.

The configuration name currently suggests an exclusive `local`/`oidc` choice even though OIDC mode retains local login. Either that mixed behavior should be documented explicitly or the option should be renamed to reflect that OIDC is an additional authentication method.

Additional implementation details worth addressing:

- Include `redirect_uri` in `isConfigured()` and declare the cURL extension dependency.
- Use the existing language catalogue for user-visible button labels and authentication errors.
- Make provisioning atomic or handle duplicate-account races safely.
- Add automated coverage beyond the current same-site Keycloak/PostgreSQL manual test; no CI checks are currently reported for this branch.

Overall, this is useful for a multidomain installation backed by one common identity tenant, or for a hybrid setup with one OIDC tenant plus local accounts. It is not yet a general multidomain/multitenant OIDC solution, and the cross-site callback behavior should be verified with Microsoft Entra, Google, and other externally hosted providers before claiming compatibility.

doneisgood:
Thank you for the thorough review. I've addressed items I could address and documented what I could not.

## Changes Made

### Security Fixes
- **Issuer validation** — ID token `iss` claim must match configured issuer URL
- **Audience validation** — ID token `aud` claim must match configured client ID (handles both string and array formats per OIDC spec)
- **Nonce required** — changed from optional to mandatory (must be present and match)
- **Active account check** — fixed broken check for disabled admin accounts (now queries `active` column directly)
- **Discovery document issuer** — validates `.well-known/openid-configuration` document's `issuer` matches configured issuer
- **UserInfo sub verification** — verifies UserInfo `sub` matches ID token `sub` before merging claims

### Features
- **MFA enforcement** — new `$CONF['oidc_mfa']` config with three modes: `'none'` (default), `'totp_fallback'` (use local TOTP if IdP lacks MFA), `'required'` (MFA mandatory via IdP or TOTP fallback). Reads `amr` claim from IdP token to detect MFA at provider. I'd appreciate feedback on whether this approach is straightforward enough or needs simplification.
- **Atomic provisioning** — uses database-specific upsert (`ON CONFLICT DO NOTHING` / `INSERT IGNORE`) to eliminate race condition on concurrent first-logins
- **Language catalogue** — SSO button text uses `Config::lang('pLogin_oidc_button')` with config override
- **cURL dependency** — declared in composer.json

### API Changes
- **Config renamed** — `$CONF['auth_provider']` → `$CONF['additional_auth']` (array). Allows multiple additional auth methods (future: LDAP, SAML). Local password auth always available.

## Known Issues (Documented)
- SameSite=Strict session cookie (cross-site callbacks)
- Identity binding by email instead of `issuer + sub` (requires schema coordination)
- ~~No discovery document issuer validation~~ (fixed)
- ~~UserInfo `sub` mismatch not verified~~ (fixed)
- **No per-domain/multitenant OIDC** — One global OIDC provider. Per-domain OIDC could benefit from further discussion on design and use cases.

## Deviations Documented
- Atomic upsert instead of upstream's check-then-insert pattern (rationale + trade-off explained)

## Not Addressed
- **Automated CI tests** — upstream test infrastructure; would benefit from a separate discussion on mocking strategy for OIDC providers
- **Identity binding by email** — requires schema changes; prefer to coordinate with upstream on the design

TrapoSAMA:
Thanks for the detailed response and the substantial follow-up work. The issuer, audience, nonce, discovery, UserInfo, inactive-account, configuration, localization, and dependency changes address most of my original concerns.

For the initially stated Keycloak/Auth0/Okta scope, I think the remaining priorities are:

- Resolve or test the `SameSite=Strict` callback behavior. The same-site Keycloak deployment may work, but standard Auth0 and Okta hosted login domains normally produce a cross-site callback.
- Exercise the actual `OIDC::handleCallback()` and provisioning paths in tests. The current tests validate JWT decoding and duplicate some comparisons, but do not verify the production flow.
- Correct or explicitly limit the database-specific provisioning SQL. The PostgreSQL/SQLite branch uses `now()`, which is not valid SQLite syntax, despite the documented SQLite test result.
- Clarify the MFA modes: the current code still invokes local TOTP when `oidc_mfa = 'none'`, so that mode does not behave as documented.
- If auto-provisioning remains enabled by default, define a policy for unverified email claims; otherwise, defaulting auto-provisioning to disabled would be safer.

With those points addressed, this would look like a reasonable first implementation for a single OIDC provider, while per-domain and multitenant OIDC can remain future work.

doneisgood:
## Update: Follow-up items addressed

Thank you for the follow-up. I've addressed all 5 items:

### 1. SameSite=Strict callback behavior
Added `$CONF['oidc_cookie_samesite']` config (default: `'Strict'`). Set to `'Lax'` when using external IdPs (Auth0, Okta, Google, Microsoft Entra). This is documented as a known issue with the config workaround.

### 2. Integration tests for `handleCallback()`
Added `tests/OIDCIntegrationTest.php` — 13 tests that exercise the full `handleCallback()` flow with a mock HTTP client:
- Successful callback (discovery → JWKS → token → UserInfo)
- State mismatch / missing state
- Discovery failure / JWKS failure / token failure
- Issuer mismatch / audience mismatch
- nonce mismatch / missing nonce
- Expired token rejection
- UserInfo `sub` mismatch

All 13 integration tests + 12 existing unit tests = 25 tests passing locally.

### 3. SQLite `now()` syntax in provisioning
Fixed: `now()` → `CURRENT_TIMESTAMP` (valid across PostgreSQL, SQLite, MySQL).

### 4. MFA modes corrected
Fixed: `oidc_mfa='none'` no longer invokes TOTP fallback. Renamed modes to be clearer:

| Mode | Behavior |
|------|----------|
| `'none'` | No MFA. TOTP not invoked. |
| `'mfa_or_totp'` | IdP MFA or local TOTP required. Rejects if neither. |
| `'idp_mfa'` | IdP MFA only. TOTP is NOT a fallback. |

### 5. Auto-provisioning + unverified email policy
- Changed `oidc_auto_provision` default to `false` (safer out of the box)
- Added `$CONF['oidc_require_verified_email']` config to enforce `email_verified` claim from IdP

## Summary of all changes since last update

| File | Change |
|------|--------|
| `config.inc.php` | Added `oidc_cookie_samesite`, `oidc_require_verified_email`, changed `oidc_auto_provision` default to `false`, renamed MFA modes |
| `public/oidc_callback.php` | Fixed SQLite timestamps, fixed MFA logic, added verified email check |
| `model/OIDC.php` | Consolidated security logging into `logSecurityEvent()` with client IP |
| `tests/OIDCIntegrationTest.php` | NEW: 13 integration tests for full callback flow |
| `DOCUMENTS/OIDC-Feature.md` | Updated docs |

All changes deployed to test environment and verified (login, SSO button, OIDC redirect all PASS).

doneisgood:
## PR Status

This PR is ready for review. All review items have hopefully been addressed:

### Implemented
- Global OIDC support (single IdP)
- Security: iss/aud/nonce validation, discovery issuer check, UserInfo sub verification, CSRF/state protection
- MFA enforcement with TOTP fallback (`none`/`mfa_or_totp`/`idp_mfa`)
- amr method whitelist/blacklist for fine-grained MFA control
- Atomic provisioning, language catalogue, cURL dependency
- 34 tests (12 unit + 22 integration), all passing

### Future Work (separate PR)
- **Per-domain OIDC** — different IdPs per domain, domain-scoped auto-provisioning, master admin vs domain admin roles. Design doc attached (`DOCUMENTS/Per-Domain-OIDC-Design.md`). Requires schema changes (domain_oidc table) and UI work. Out of scope for this PR.
- **Identity binding by issuer+sub** — more stable than email. Requires schema coordination.

### Concerns
- **MFA Logic** — This logic may have become over-engineered and gotten away from me more than once
- **Testing** — All testing was done with keycloak.  More OIDC implementations need tested
- **Design** — This was designed for need.  I hope it addresses more than just my need.

doneisgood:
## Update: SQLite provisioning tests added

Added `tests/OIDCSQLiteProvisioningTest.php` — 4 tests that verify the actual provisioning SQL works against SQLite:

- `INSERT ... CURRENT_TIMESTAMP` works
- `ON CONFLICT DO NOTHING` is idempotent
- `CURRENT_TIMESTAMP` produces valid datetime
- Concurrent inserts don't create duplicates

This addresses the reviewer's concern about SQLite compatibility — we now have actual tests proving the provisioning SQL works, not just assuming it does.

### Testing strategy for future work

The SQLite test infrastructure is now established. For the per-domain OIDC PR (future work), we'll build on this pattern:

1. **Test setup** — create table, clean state
2. **Test the actual SQL** — same queries from the callback
3. **Test edge cases** — idempotency, concurrency, valid data
4. **Test teardown** — clean up

This will cover:
- `domain_oidc` table operations (INSERT, UPDATE, DELETE per-domain configs)
- Domain-scoped provisioning (user gets `domain_admins` entry, not super-admin)
- Domain-to-IdP routing (which IdP config applies to which domain)

### Total test count: 38

| Test File | Tests | Coverage |
|-----------|-------|----------|
| `OIDCTest.php` | 12 | JWT decoding, amr claims |
| `OIDCIntegrationTest.php` | 22 | Full callback flow with mock HTTP |
| `OIDCSQLiteProvisioningTest.php` | 4 | Actual SQLite DB operations |

doneisgood:
## Update: Auto-provisioning flow tests added

Added `tests/OIDCAutoProvisioningTest.php` — 5 tests that verify the full auto-provisioning flow with SQLite:

- Full flow creates admin user (handleCallback → INSERT → verify)
- Auto-provisioning disabled rejects new user
- Existing user not affected by auto-provisioning
- Disabled user cannot login
- Verified email requirement blocks unverified

### Total test count: 43

| Test File | Tests | Coverage |
|-----------|-------|----------|
| `OIDCTest.php` | 12 | JWT decoding, amr claims |
| `OIDCIntegrationTest.php` | 22 | Full callback flow with mock HTTP |
| `OIDCSQLiteProvisioningTest.php` | 4 | SQLite provisioning SQL |
| `OIDCAutoProvisioningTest.php` | 5 | Full auto-provisioning flow |

config.inc.php
```diff
@@ -807,6 +807,47 @@ function x_send_sms_clickatell($to, $message) {
 
 $CONF['version'] = '4.0.1';
 
+// Local password auth is always available.
+// Additional authentication methods can be enabled here.
+// Example: ['oidc'] or ['oidc', 'ldap'].
```
DavidGoodwin: can you change the comment so it doesn't reference ldap (which doesn't exist in this codebase yet?)

doneisgood: The documentation mentioned that additional methods were aspirational .. but this might mislead someone into thinking that ldap existed.   I'll get that on the list for the next update to this PR.   Thank you!

[Ed: I swear that I had it documented, but it's completely not there now]

doneisgood:
Oops.. posted to the wrong thread (you can find it at #1145 now)

config.inc.php
doneisgood: LDAP stuff is designed in a different branch... removed here.

PR #1146, opened 2026-09-04
Show inactive domain DNS status in overview

TrapoSAMA:
## Summary

- Cache binary domain DNS status without DNS lookups during page rendering.
- Show a conditional `DNS (N)` alert/filter and an on-demand refresh beside the administrator selector; retain only the accessible warning icon in each affected domain row.
- Support `domain_dns_status_check`: `0` disabled, `1` authoritative zone, `2` usable MX.
- Show the last completed batch time and each domain's own status/check time beside its selector, with individual refresh.
- Document installation, an optional CLI/cron command, permissions, networking, timeouts, and saved-result semantics.
- Explain the basic hourly crontab without requiring `flock`, then explain the optional utility, `-n`, lock-file permissions, and its limitations separately. Markdown delimiters are not commands.
- Declare `dns_checked` as `timestamp NULL DEFAULT NULL` to avoid MariaDB error 1067 with legacy implicit TIMESTAMP nullability.

## DNS checks and refresh scope

Zone mode requires NS delegation and at least one authoritative server answering a direct SOA query successfully. MX mode requires an MX target with a resolvable A or AAAA address. No WARNING or advanced diagnostic states are introduced.

The CLI checks all domains. Web batch refresh is restricted to the selected administrator's authorized domain set. Completing the same set updates the same batch timestamp; individual checks do not advance it. Interrupted batches preserve completed individual results without recording batch completion. Opening a page never forces a DNS refresh. Checks remain sequential; cron is optional and large manual batches can hit web request timeouts.

Thanks to @doneisgood for pointing out that cached results need visible check times.

## Migration coordination

Integrated `master` at e6a0b3a179ad8a239c6a4522bda1fc9937582788 (schema 1856); the DNS migration is provisionally 1857. PR #1054 independently uses the same next number. Whichever merges second must be rebased and renumbered. The retained production sequence is broadcast 1857 followed by DNS 1858. Batch metadata reuses the existing config table and needs no additional migration.

## Validation

- PHP 8.4/SQLite focused suite: 13 tests, 50 assertions, including rendered Smarty headers.
- Focused Psalm, formatting, and language-catalog checks.
- Linux CI run [34008492992](https://github.com/postfixadmin/postfixadmin/actions/runs/34008492992) passed lint/Psalm, PHP 8.2–8.5, MySQL, and PostgreSQL on the preceding implementation revision; it does not cover the latest documentation/nullability follow-up.
- The explicit-nullability correction was verified by completing the interrupted local MariaDB upgrade from schema 1857 to 1858 in the production patch variant, followed by direct authenticated page checks. No additional local suites were run for this follow-up.
- The retained production patch passed GNU `patch -p1` dry-run and actual application without offsets, fuzz, or rejects; the applied tree exactly matches its production-specific source.
- Full local suite attempted: unrelated failures involving Unix commands/Windows paths, environment DNS behavior, and missing GD; not reported as passing.
- Existing refresh/filter/count behavior was tested by the project owner on a real installation; the newly added timestamp/individual controls still need their production UI review.

The search-field label was separately merged in #1148. Conflicts are resolved in 7b11e5c4f6a0730893c83332ce167fb435bc389b, preserving both its translations/partial and the DNS controls. The focused integration checks passed: 14 tests, 59 assertions on PHP 8.4/SQLite. Updated CI runs on this merge; no XAMPP installation changes were made for this integration.

### DIFF
--- a/model/DomainDnsStatus.php
+++ b/model/DomainDnsStatus.php
@@ -1,0 +1,191 @@
+<?php
+
+/**
+ * Checks and stores the binary DNS status shown in the domain overview.
+ *
+ * Lookups are performed only by refresh(), never while rendering a list.
+ */
+class DomainDnsStatus
+{
+    private int $mode;
+
+    public function __construct(private float $timeout = 1.0, ?int $mode = null)
+    {
+        $this->mode = $mode ?? self::configuredMode();
+    }
+
+    public static function configuredMode(): int
+    {
+        $configured = Config::read('domain_dns_status_check');
+        if (!is_int($configured) && !is_string($configured)) {
+            return 0;
+        }
+        $mode = (int)$configured;
+        return in_array($mode, [0, 1, 2], true) ? $mode : 0;
+    }
+
+    /** @param string[] $domains
+     *  @return array{active: int, inactive: int}
+     */
+    public function refresh(array $domains): array
+    {
+        $result = ['active' => 0, 'inactive' => 0];
+        if ($this->mode === 0) {
+            return $result;
+        }
+        foreach (array_unique($domains) as $domain) {
+            if (!is_string($domain) || $domain === '' || $domain === 'ALL') {
+                continue;
+            }
+            $active = $this->isActive($domain);
+            db_update('domain', 'domain', $domain, [
+                'dns_active' => $active,
+                'dns_checked' => date('Y-m-d H:i:s'),
+            ], []);
+            $result[$active ? 'active' : 'inactive']++;
+        }
+        return $result;
+    }
+
+    public function isActive(string $domain): bool
+    {
+        if ($this->mode === 2) {
+            return $this->hasUsableMx($domain);
+        }
+        if ($this->mode !== 1) {
+            return false;
+        }
+        foreach ($this->nameservers($domain) as $nameserver) {
+            foreach ($this->nameserverAddresses($nameserver) as $address) {
+                if ($this->authoritativeServerResponds($domain, $address)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    protected function hasUsableMx(string $domain): bool
+    {
+        foreach ($this->mxTargets($domain) as $target) {
+            if ($this->nameserverAddresses($target) !== []) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /** @return string[] */
+    protected function mxTargets(string $domain): array
+    {
+        if (!function_exists('dns_get_record')) {
+            return [];
+        }
+        $records = @dns_get_record(rtrim($domain, '.'), DNS_MX);
+        if (!is_array($records)) {
+            return [];
+        }
+        $targets = [];
+        foreach ($records as $record) {
+            if (isset($record['target']) && is_string($record['target'])) {
+                $target = strtolower(rtrim($record['target'], '.'));
+                if ($target !== '') {
+                    $targets[] = $target;
+                }
+            }
+        }
+        return array_values(array_unique($targets));
+    }
+
+    /** @param string[] $domains */
+    public static function countInactive(array $domains): int
+    {
+        if ($domains === []) {
+            return 0;
+        }
+        $params = ['dns_active' => false];
+        $where = db_in_clause('domain', $domains, $params);
+        $table = table_by_key('domain');
+        $row = db_query_one("SELECT count(*) AS inactive_count FROM $table WHERE dns_active = :dns_active AND $where", $params);
+        return (int)($row['inactive_count'] ?? 0);
+    }
+
+    /** @return string[] */
+    protected function nameservers(string $domain): array
+    {
+        if (!function_exists('dns_get_record')) {
+            return [];
+        }
+        $records = @dns_get_record(rtrim($domain, '.'), DNS_NS);
+        if (!is_array($records)) {
+            return [];
+        }
+        $nameservers = [];
+        foreach ($records as $record) {
+            if (isset($record['target']) && is_string($record['target'])) {
+                $nameservers[] = strtolower(rtrim($record['target'], '.'));
+            }
+        }
+        return array_values(array_unique($nameservers));
+    }
+
+    /** @return string[] */
+    protected function nameserverAddresses(string $nameserver): array
+    {
+        $records = @dns_get_record($nameserver, DNS_A | DNS_AAAA);
+        if (!is_array($records)) {
+            return [];
+        }
+        $addresses = [];
+        foreach ($records as $record) {
+            if (isset($record['ip']) && filter_var($record['ip'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV4)) {
+                $addresses[] = $record['ip'];
+            } elseif (isset($record['ipv6']) && filter_var($record['ipv6'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
+                $addresses[] = $record['ipv6'];
+            }
+        }
+        return array_values(array_unique($addresses));
+    }
+
+    protected function authoritativeServerResponds(string $domain, string $address): bool
+    {
+        $id = random_int(0, 65535);
+        $query = pack('nnnnnn', $id, 0, 1, 0, 0, 0) . $this->encodeName($domain) . pack('nn', 6, 1);
+        $endpoint = str_contains($address, ':') ? "udp://[$address]:53" : "udp://$address:53";
+        $socket = @stream_socket_client($endpoint, $errno, $error, $this->timeout, STREAM_CLIENT_CONNECT);
+        if ($socket === false) {
+            return false;
+        }
+        $seconds = (int)$this->timeout;
+        $microseconds = (int)round(($this->timeout - (float)$seconds) * 1000000.0);
+        stream_set_timeout($socket, $seconds, $microseconds);
+        $written = @fwrite($socket, $query);
+        $response = $written === strlen($query) ? @fread($socket, 4096) : false;
+        fclose($socket);
+        if (!is_string($response) || strlen($response) < 12) {
+            return false;
+        }
+        $header = unpack('nid/nflags', substr($response, 0, 4));
+        if (!is_array($header)) {
+            return false;
+        }
+        $flags = (int)$header['flags'];
+        return (int)$header['id'] === $id
+            && ($flags & 0x8000) !== 0
+            && ($flags & 0x0400) !== 0
+            && ($flags & 0x000f) === 0;
+    }
+
+    private function encodeName(string $domain): string
+    {
+        $encoded = '';
+        foreach (explode('.', rtrim($domain, '.')) as $label) {
+            $length = strlen($label);
+            if ($length === 0 || $length > 63) {
+                return "\0";
+            }
+            $encoded .= chr($length) . $label;
+        }
+        return $encoded . "\0";
+    }
+}
--- a/model/DomainHandler.php
+++ b/model/DomainHandler.php
@@ -113,6 +113,7 @@ class DomainHandler extends PFAHandler
             #                           editing?    form    list
            'domain'            => self::pacol($this->new, 1,      1,      'text', 'domain'                       , ''                                 ,'', array(), 0, 0, "", "", 'list-virtual.php?domain=%s'),
            'full_mailbox_count' => self::pacol(0,          0,      1,      'vnum', ''                             , ''                                 , '', array(), 0, 0, 'coalesce(__full_mailbox_count,0) as full_mailbox_count'),
+           'dns_active'        => self::pacol(0,          0,      1,      'bool', ''                             , ''),
            'description'       => self::pacol($super,     $super, $super, 'text', 'description'                  , ''),
 
            # Aliases
--- a/public/list.php
+++ b/public/list.php
@@ -89,7 +89,21 @@ if (array_key_exists('reset_search', $_GET)) {
 $_SESSION["search_$table"] = $search;
 $_SESSION["searchmode_$table"] = $searchmode;
 
-$condition = count($search) ? $search : '';
+$dns_filter = '';
+$dns_inactive_count = 0;
+$dns_check_mode = DomainDnsStatus::configuredMode();
+if ($table === 'domain' && $dns_check_mode > 0) {
+    $dns_filter = safeget('dns_filter') === 'inactive' ? 'inactive' : '';
+    $dns_inactive_count = DomainDnsStatus::countInactive(list_domains_for_admin($username));
+}
+
+$condition = $search;
+if ($dns_filter === 'inactive') {
+    $condition['dns_active'] = 0;
+}
+if ($condition === []) {
+    $condition = '';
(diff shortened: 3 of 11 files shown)


doneisgood:
Thanks for the PR — caching DNS status is a great addition. One concern:

**The cached value never expires.** If a domain's DNS breaks 5 minutes after a manual refresh, the UI will show "active" until someone clicks refresh again. The `dns_checked` timestamp is stored but not used for anything, so there's no way to know how stale the data is.

**Suggestion:** Respect the DNS TTL. `dns_get_record()` already returns TTL per record. Could we:

1. Store the TTL alongside `dns_active` in `domain` table
2. Compare `dns_checked + dns_ttl` to current time to detect staleness
3. Either:
   - Show a staleness indicator ("checked 3h ago, TTL was 5m")
   - Do a lazy refresh: serve cached value immediately, trigger background refresh so next page load is fresh

This keeps the "no lookup on render" win while preventing the false-confidence problem.

~~The lazy refresh approach is cleanest — the user never waits, and the cache is always fresh on the next request.~~
[Ed:  I reread this PR, if the user knows staleness.. they can trigger their own refresh and not have 'stuff' done on their behalf.]

TrapoSAMA:
Thanks @doneisgood — I agree that a saved result should not look like a live check.

Updated in 333a359b6fe73ccbbdb2a9a29fe69c27ebc74103:

- The domain overview now shows the last completed check for the selected administrator's domain set. The optional global CLI/cron check and a manual refresh of that same set update the same batch timestamp.
- Each domain's virtual-address page shows its own saved status and check time beside the domain selector, with an individual refresh button. Checking one domain does not advance the batch timestamp.
- Unchecked domains are identified as such. An interrupted batch retains completed individual results without advancing the batch timestamp. Authorization and CSRF checks apply to individual refreshes too.
- The documentation now includes a concrete hourly crontab example, service-account requirements, an optional flock guard, and the sequential-check/request-timeout limitations. Cron remains optional.

These are explicit health-check timestamps, not DNS record TTLs. Page rendering remains read-only: no DNS lookup, automatic refresh, or additional WARNING state. Users can see when the saved result was obtained and choose when to refresh it.

The focused PHP 8.4/SQLite tests pass (13 tests, 50 assertions), including rendered Smarty headers, authorization, disabled mode, interrupted batches, and independent individual/batch timestamps. Focused Psalm and language-catalog checks pass. The full local suite has unrelated Windows/Unix-command, DNS-environment, and missing-GD failures. [Linux CI](https://github.com/postfixadmin/postfixadmin/actions/runs/34008492992) has now passed lint/Psalm, PHP 8.2–8.5, MySQL, and PostgreSQL on this revision.

TrapoSAMA:
Documentation follow-up: the basic cron example now has no `flock` dependency. It explains the schedule fields, `crontab -e`, account permissions, real executable/application paths, and that Markdown fences are not commands. A separate optional section introduces `flock`, explains `-n` with an overlapping-run example, and states that it does not serialize web refreshes. Manual refresh remains sufficient without cron.

Also included the local installation correction for MariaDB error 1067: explicitly nullable `timestamp NULL DEFAULT NULL`. The interrupted upgrade subsequently completed successfully on local MariaDB, and both affected authenticated pages opened normally. No new local test suites were run for this follow-up.

The merged search-label PR #1148 causes conflicts with current master; integration remains pending rather than being included silently in this follow-up.

TrapoSAMA:
Resolved the conflicts with master in 7b11e5c4f6a0730893c83332ce167fb435bc389b. Both the DNS controls and the search label/partial from #1148 are retained, along with all language keys. Updated the DNS rendering test to load the merged search partial.

Focused integration checks passed: 14 tests, 59 assertions on PHP 8.4/SQLite. The documentation and explicit nullable timestamp correction are retained. Updated CI runs on this revision.

public/list.php
```diff
@@ -89,7 +89,23 @@
 $_SESSION["search_$table"] = $search;
 $_SESSION["searchmode_$table"] = $searchmode;
 
-$condition = count($search) ? $search : '';
+$dns_filter = '';
+$dns_inactive_count = 0;
+$dns_group_checked = '';
+$dns_check_mode = DomainDnsStatus::configuredMode();
+if ($table === 'domain' && $dns_check_mode > 0) {
```
cboltz, line 95: Needing such a table-specific condition in list.php is a sign that you are holding it wrong ;-)

You probably want to use `?search[dns_active]=0` (where `dns_filter` is the field you added to DomainHandler), which should give you all the handling for free.

model/DomainHandler.php
```diff
@@ -113,6 +113,7 @@ protected function initStruct()
             #                           editing?    form    list
            'domain'            => self::pacol($this->new, 1,      1,      'text', 'domain'                       , ''                                 ,'', array(), 0, 0, "", "", 'list-virtual.php?domain=%s'),
            'full_mailbox_count' => self::pacol(0,          0,      1,      'vnum', ''                             , ''                                 , '', array(), 0, 0, 'coalesce(__full_mailbox_count,0) as full_mailbox_count'),
+           'dns_active'        => self::pacol(0,          0,      1,      'bool', ''                             , ''),
```
cboltz, line 116: In upgrade.php, you also add `dns_checked`. This should also be listed here.

model/DomainDnsStatus.php
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /** Only expose status for a domain in the caller's authorized scope. */
+    public static function domainStatus(string $domain, array $allowedDomains): array
+    {
+        if (!in_array($domain, $allowedDomains, true)) {
+            throw new InvalidArgumentException('Domain outside administrator scope');
+        }
+        $table = table_by_key('domain');
+        return db_query_one("SELECT dns_active, dns_checked FROM $table WHERE domain = :domain", ['domain' => $domain]) ?? [];
```
cboltz: With `dns_checked` also added to DomainHandler $struct, this whole function looks superfluous - and you'll get the $allowedDomains check etc. for free in DomainHandler.

Actually I wonder if the most of, or even the whole class {c,sh}ould be integrated into DomainHandler...

model/DomainDnsStatus.php
```diff
@@ -0,0 +1,233 @@
...
+    public function refreshGroup(array $domains): array
+    {
+        $result = $this->refresh($domains);
+        if ($this->mode !== 0) {
+            $table = table_by_key('config');
+            $sql = "INSERT INTO $table (name, value) VALUES (:name, :value)";
+            $sql .= db_mysql()
+                ? ' ON DUPLICATE KEY UPDATE value = VALUES(value)'
+                : ' ON CONFLICT (name) DO UPDATE SET value = excluded.value';
+            db_query($sql, ['name' => self::groupKey($domains), 'value' => date('Y-m-d H:i:s')]);
```
cboltz: So you record the date of the last check
-  in the domain table and
- also in the config table - which looks superfluous, and is a candidate for drifting from the domain table

Better record this only in the domain table, and check for the oldest date in a group of domains (or simply check for `date < fresh_enough_date`)

cboltz:
Note that I only did a quick review, so I might find more issues when I have some more time ;-)

TrapoSAMA:
Thanks Chris! I had to leave a few easter eggs for you to find — can't get everything perfect on the first try 😄

All good points. I'll go through them and clean things up accordingly.

public/list.php
TrapoSAMA: Fixed in dd4642fa. I removed the table-specific dns_filter branch from list.php and the template. The alert now uses search[dns_active]=0, so filtering, session persistence and pagination go through the existing list machinery; clearing it uses reset_search=1.

model/DomainDnsStatus.php
TrapoSAMA: Fixed in dd4642fa. Saved-status reads and writes now go through DomainHandler::dnsStatus() and updateDnsStatus(), so DomainHandler::init() enforces the caller's domain scope. I also moved the scoped inactive count and oldest-check query into the handler. DomainDnsStatus now remains responsible only for DNS network probing and check orchestration; keeping socket/DNS protocol work out of the persistence handler preserves that separation.

model/DomainDnsStatus.php
TrapoSAMA: Fixed in dd4642fa. The config-table batch timestamp and its upsert/key logic were removed. The overview now derives one value from domain.dns_checked: the oldest timestamp in the authorized domain set, but only when every domain has been checked; otherwise it shows Not checked. This removes the possible drift and also makes partial or interrupted runs visible.

model/DomainHandler.php
TrapoSAMA: Fixed in dd4642fa. dns_checked is now part of the DomainHandler struct as a read-only timestamp field. It is selected for handler reads without adding another visible list column.

TrapoSAMA: Follow-up in 91bfbcd3: MySQL and PostgreSQL CI showed that the handler's inherited empty-string default was being inserted into the timestamp column for newly created domains. The dns_checked struct default is now explicitly null, matching the nullable migration and the unchecked state. A regression assertion covers this default.

TrapoSAMA:
Thanks for the review. The four points are now addressed:

- The inactive alert uses the standard `search[dns_active]=0` list filter; the special `dns_filter` path was removed.
- `dns_checked` is declared in `DomainHandler` with an explicit `NULL` default, matching the nullable schema and the unchecked state.
- Saved-status reads, writes, scoped counts, and the oldest-check query now go through `DomainHandler`, including its domain-scope enforcement. `DomainDnsStatus` is limited to DNS network probing and orchestration.
- Duplicate batch timestamps in the config table were removed. The overview derives the oldest saved domain check and reports `Not checked` if any domain in scope remains unchecked, making partial runs visible without drift.

The initial follow-up exposed the timestamp empty-string default in MySQL/PostgreSQL CI; commit `91bfbcd3` corrects it and adds a regression assertion. GitHub Actions run 34218435748 now passes lint/static analysis, PHP 8.2-8.5, MySQL, and PostgreSQL.

model/DomainDnsStatus.php
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    public static function configuredMode(): int
+    {
+        $configured = Config::read('domain_dns_status_check');
+        if (!is_int($configured) && !is_string($configured)) {
+            return 0;
+        }
+        $mode = (int)$configured;
+        return in_array($mode, [0, 1, 2], true) ? $mode : 0;
```
DavidGoodwin: can you make 0 , 1 + 2 constants or something within this class, and then do comparisons against them? 

e.g. public const CHECK_DISABLED = 0; public const CHECK_ZONE = 1 , public const CHECK_MX =2  ... or something similar?

(As postifxadmin is supporting PHP 8.2 and above, you could use a proper enum if you'd rather).

model/DomainDnsStatus.php
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /** @param string[] $domains
+     *  @return array{active: int, inactive: int}
+     */
+    public function refresh(array $domains): array
+    {
+        $result = ['active' => 0, 'inactive' => 0];
+        if ($this->mode === 0) {
```
DavidGoodwin: constant comparison or something please....

model/DomainDnsStatus.php
```diff
@@ -0,0 +1,177 @@
...
+                $addresses[] = $record['ip'];
+            } elseif (isset($record['ipv6']) && filter_var($record['ipv6'], FILTER_VALIDATE_IP, FILTER_FLAG_IPV6)) {
+                $addresses[] = $record['ipv6'];
+            }
+        }
+        return array_values(array_unique($addresses));
+    }
+
+    protected function authoritativeServerResponds(string $domain, string $address): bool
+    {
```
DavidGoodwin, line 148: i could do with a comment here explaining why the code is doing all the packing / socket stuff. 


is this code just trying to verify whether the DNS server for the domain (in postfixadmin) exists/responds?

Is it really worthwhile performing this sort of check? I get looking to see whether a domain resolves and has an MX record ... is there any value in going further?

Perhaps I'm not understanding this PR enough yet.

model/DomainHandler.php
```diff
@@ -165,6 +167,59 @@ protected function initStruct()
...
+    /** Return the saved DNS result for a domain in this handler's scope. */
+    public function dnsStatus(string $domain): array
+    {
+        if (!$this->init($domain)) {
+            throw new InvalidArgumentException('Domain outside administrator scope');
+        }
+        $active = $this->result['dns_active'] ?? null;
+        return [
+            'dns_active' => $active === null ? null : (int)$active,
+            'dns_checked' => $this->result['_dns_checked'] ?? null,
```
DavidGoodwin, line 193: why _dns_checked here?

model/DomainDnsStatus.php
TrapoSAMA: Thanks — fixed in 76694b25. The three configuration values are now exposed as CHECK_DISABLED, CHECK_ZONE and CHECK_MX constants. I kept them as integer constants because the public configuration value remains an integer; an enum would add conversion without improving the boundary here.

model/DomainDnsStatus.php
TrapoSAMA: Fixed in 76694b25. configuredMode(), refresh() and isActive() now compare against the named mode constants instead of 0, 1 and 2.

model/DomainDnsStatus.php
TrapoSAMA: Thanks — I added a focused comment in 76694b25. Zone mode deliberately sends an SOA query directly to each delegated nameserver and requires a matching, authoritative, successful DNS response. A recursive NS lookup only proves that delegation records exist (and may be cached); it does not prove that an authoritative server is reachable and serving the zone. This implements the zone-health mode requested by the feature. MX mode remains the simpler alternative for installations that only want a usable mail-routing check.

model/DomainHandler.php
TrapoSAMA: PFAHandler exposes timestamp fields twice: dns_checked is formatted through the general localized date format, which contains only the date, while _dns_checked is the raw stored timestamp. This UI needs both date and time for the actual check instant, so the raw alias is intentional. I added an inline explanation in 76694b25 rather than changing the established handler-wide timestamp formatting behavior. Focused validation passes 13 tests and 50 assertions.

DavidGoodwin:
this page looks fine - and when i hover over the little warning triangle I see e.g. "Inactive DNS".

image

when I go to a specific domain, i still see the warning triangle, but there is no hover text (and when viewing a specific domain I'd expect the "Inactive DNS" to be visible without needing to hover my mouse over something.

image

TrapoSAMA:
Thanks — fixed in `9fb89637`.

The selected-domain view now renders the warning icon together with the visible localized `Inactive DNS` text. The icon is decorative (`aria-hidden`) because the adjacent text provides the accessible status.

No status indicator is rendered when DNS is active; the page remains quiet unless attention is required. The focused Smarty rendering test now covers inactive, active, and unchecked states (2 tests, 25 assertions).

PR #1147, opened 2026-09-06
Per-Domain OIDC: schema, migration, handler, callback, login flow

doneisgood:
## Summary

Adds per-domain OIDC support on top of PR #1144 (global OIDC).

- Add `domain_oidc` table for per-domain IdP configuration
- Add `oidc_issuer` and `oidc_sub` columns to `admin` table for stable identity binding
- Add `oidc_identity` config switch: `email` (legacy) or `issuer_sub` (recommended)
- Add `oidc_upgrade_existing` config: upgrade NULL accounts with issuer+sub on first OIDC login
- Add `DomainOidcHandler` model for CRUD operations
- Update callback for per-domain routing + issuer+sub identity binding
- Update login flow for per-domain OIDC initiation
- Add UNIQUE constraint on `domain_admins(username, domain)` to prevent duplicates
- Add tests for per-domain OIDC

## Dependencies

This PR depends on PR #1144 (global OIDC support). It should be reviewed/merged after #1144.

## Related

- GitHub Discussion #1145
- PR #1144 (global OIDC support)
- PR #1146 (DNS status — uses upgrade_1857, we use 1859/1860 to avoid collision)

## Testing

- All 59 tests pass on LXC 114
- CI green on doneisgood/postfixadmin (lint_etc, testsuite, testsuite-mysql, testsuite-pgsql)
- Tested end-to-end with two Keycloak realms


doneisgood:
Sorry, upon extra review and testing:
* ~~I didn't get a UI in place for the Per-Domain managed OIDC stuff to edit what's in the database.~~ (fixed)
* ~~The Add Domain doesn't have labels on the OIDC stuff.~~  (fixed)

Not calling this feature complete without it.

tests/bootstrap.php
```diff
@@ -24,6 +24,20 @@
     touch($db_file);
 
     error_log("Using: SQLite database for tests - $db_file");
+
```
DavidGoodwin, line 27: why is this db table addition needed?

tests/DomainOidcHandlerTest.php
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        // Reset global OIDC config to known state
+        global $CONF;
+        $CONF['oidc_mfa'] = 'none';
+        $CONF['oidc_mfa_methods'] = ['mfa', 'otp', 'hwk'];
+        $CONF['oidc_mfa_blacklist'] = [];
+
+        // Create tables
+        db_execute("CREATE TABLE IF NOT EXISTS domain (
```
DavidGoodwin, line 26: i'm not sure the unit tests need to create tables - that should be done by bootstrap.php

model/DomainOidcHandler.php
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    /**
+     * Get OIDC config for this domain
+     * @return array|null Config array or null if not configured
+     */
+    public function get(): ?array
+    {
+        $table = table_by_key('domain_oidc');
+        $config = db_query_one("SELECT * FROM $table WHERE domain = ?", [$this->domain]);
```
cboltz, line 26: This is just a very quick review (without reviewing the whole PR), but I have some questions:
- what's the reason to have a separate domain_oidc table instead of adding fields to the domain table?
- what's the reason for the separate DomainOidcHandler instead of adding the handling to DomainHandler?
- on a more generic level - DomainOidcHandler isn't based on PFAHandler, and therefore shouldn't have a name that ends in *Handler

tests/bootstrap.php
doneisgood: OK.  removed that.. made sure the tables were created in bootstrap.  I needed clear testing in my envirionment.  oops.

tests/DomainOidcHandlerTest.php
doneisgood: they should *not*.   addressed.

doneisgood:
Thanks for the review, cboltz.

**Q1: Why a separate `domain_oidc` table instead of adding fields to `domain`?**

1. Zero risk to existing data — Adding 9 columns to `domain` (most NULL for non-OIDC domains) bloats a core table. New table = new code, no impact on existing deployments.
2. Follows existing pattern — `domain_admins` is already a 1:1 extension table keyed by domain with ON DELETE CASCADE.
3. Clean separation — OIDC config is a domain-lifecycle concern, not a domain-management concern.
4. Migration safety — Can be added/removed independently without ALTER TABLE on a production `domain` table.

**Q2: Why a separate `DomainOidcHandler` instead of folding into `DomainHandler`?**

1. Different lifecycle — `DomainHandler` manages domain CRUD (create/delete, password expiry, CLI, search). `DomainOidcHandler` manages OIDC config CRUD (no CLI, no search, PK-scoped, super-admin only).
2. Different permissions — Domain management is admin-scoped; OIDC config is super-admin only.
3. Implementation isolation — Allowed developing and testing the OIDC feature without risking the working domain management code.
4. Different UI path — OIDC config is a subsection of Domain Edit, not a full domain lifecycle view.

**Q3: `DomainOidcHandler` doesn't extend `PFAHandler` — name shouldn't end in `\*Handler`**

Good catch. Built standalone first to test OIDC config CRUD logic in isolation, with the plan to fold it into `PFAHandler` (proper subclass with `$db_table`, `$id_field`, etc.) before final merge. Currently a thin wrapper that doesn't use any `PFAHandler` machinery. Options:

- Rename to `DomainOidcConfig` (plain model class)
- Refactor to extend `PFAHandler` (proper upstream pattern, but more work)

Open to either approach — your preference?

Also updated the design docs (DOCUMENTS/Per-Domain-OIDC-Design.md and DOCUMENTS/Per-Domain-OIDC-Design-Updated.md) with this Q&A for reference.

Thanks!

DOCUMENTS/OIDC-Feature.md
```diff
@@ -0,0 +1,199 @@
...
+### Domain Edit Page — OIDC Section
+
+| Field | Type | Description |
+|-------|------|-------------|
+| OIDC Enabled | checkbox | Enable OIDC for this domain |
+| Issuer URL | text | IdP realm URL (e.g. `https://keycloak.example.com/realms/your-realm`) |
+| Client ID | text | OIDC client ID |
+| Client Secret | password | OIDC client secret (masked — leave blank to keep existing) |
+| Scopes | text | OAuth scopes (default: `openid email profile`) |
+| Login Button Text | text | Custom button label (default: `Login with SSO`) |
```
cboltz, line 178: Just wondering - how can the login button text be domain specific?

I'd guess that this button is shown at/before login, so we don't know the username/domain yet?

cboltz:
> Follows existing pattern — domain_admins is already a 1:1 extension table keyed by domain with ON DELETE CASCADE.

No, `domain_admins` is not a 1:1 relation to `domain`. A domain can have multiple admins, or even none (= only managed by a superadmin).

Whenever we have a 1:1 relation, we simply add fields to the existing table because that makes things much easier.

> Migration safety — Can be added/removed independently without ALTER TABLE on a production domain table.

I don't "buy" this argument. We've added quite some database fields before, and adding 9 more fields isn't any different or more risky.

> Different permissions — Domain management is admin-scoped; OIDC config is super-admin only.

Well, not reallly. Domain _management_ (creating a domain, changing the number of mailboxes etc.) is also super-admin only.

Domain admins can "only" create and edit mailboxes, aliases etc. - but they can't change the domain settings. Therefore OIDC config isn't really different IMHO.

.

For the class naming - well, so far I don't see a reason to have a separate class, and would simply add everything to DomainHandler.

> Refactor to extend PFAHandler (proper upstream pattern, but more work)

Actually less work - if you integrate everything into DomainHandler, you'll get most of the handling "for free". For example, you won't need what you added to read_from_db_postprocess() and postSave() in DomainHandler.

doneisgood:
@cboltz : It'll take me some time to rework it.  Thanks for the solid direction.

doneisgood:
Sorry about the 'force commits'  .. I've messed up the common history on my own branches and I'm not sure what the best approach is to continue this PR.   I'm working on a new PR #1158 to address the comments from *this* PR.  Maybe I'll have better thoughts later.

PR #1150, opened 2026-09-12
Clarify alias destination status indicators

TrapoSAMA:
Alias destinations absent from local records previously shared the misleading “maybe UNDELIVERABLE” indicator, including valid external addresses such as Gmail. The indicator now distinguishes “Check local destination” (red) from “External destination” (blue), with short explanatory tooltips. An alias with both kinds of destinations can show both indicators.

Destination existence is checked only within the caller's authorized domains, including when resolving an active alias domain. Targets outside that scope are classified as external without inspecting their address records. Existing catch-all, recipient-delimiter, vacation, and configured exception handling is retained. The red label requests a local-record review rather than claiming delivery must fail.

English and Spanish wording is updated. Other language catalogs retain their translations with explicit `# XXX` retranslation notes for the changed English label and untranslated markers for the new strings. Sample custom-domain colors are disabled by default, with commented examples retained.

Addresses discussion #886. Thanks @cboltz for the translation and domain-isolation review feedback.

Validation: four focused PHP 8.4/SQLite regression tests passed (15 assertions), covering mixed targets, cross-domain privacy, catch-all, recipient delimiters, exceptions, and alias-domain permissions. Focused formatting and whitespace checks completed. GitHub Actions run `34795496123` passes lint, PHP 8.2–8.5, MySQL, and PostgreSQL after correcting the test fixture's foreign-key setup. The updated production patch applied cleanly to local master, and its authenticated Spanish page displays the revised legend and tooltips. No schema or dependency changes are required.

The corresponding 4.x PR #1152 remains unchanged pending review of this master implementation.


languages/en.lang
```diff
@@ -483,7 +483,8 @@ $PALANG['pBroadcast_error_empty'] = 'The fields Name, Subject and Message should
 $PALANG['broadcast_mailboxes_only'] = 'Only send to mailboxes';
 $PALANG['broadcast_to_domains'] = 'Send to domains:';
 
-$PALANG['pStatus_undeliverable'] = 'maybe UNDELIVERABLE ';
+$PALANG['pStatus_undeliverable'] = 'Possible external destination';
```
cboltz: If you change an english text, please also add a note to all other languages to request re-translation. `language-update.sh --addcomment` does it ;-)

functions.inc.php
```diff
@@ -2442,7 +2447,9 @@ function gen_show_status($show_alias)
             }
         } // while
         if ($stat_ok == 0) {
-            $stat_string .= "<span style='background-color:" . $CONF['show_undeliverable_color'] . "'>" . $CONF['show_status_text'] . "</span>&nbsp;";
+            $stat_string .= "<span role='img' aria-label='$external_destination_tooltip'" .
+                " title='$external_destination_tooltip' style='background-color:" .
+                $CONF['show_undeliverable_color'] . "'>" . $CONF['show_status_text'] . "</span>&nbsp;";
```
cboltz: Slightly changing the text (and adding the tooltip) might be nice (even if I think that the tooltip is superfluous), but the status indicator has an unwritten TODO attached since years: Ideally it should differ between:
- local target, undeliverable for sure
- external target

Note that checking for a local target should be restricted to domains the logged in admin can access to avoid disclosing information if a mail address exists on a domain that lives in the same PostfixAdmin instance, but with a different admin.

So if you have some time, implementing the split check for local vs. external targets would be nice.

languages/en.lang
TrapoSAMA: Thanks for pointing this out. Fixed in 0f4f0bbe: the changed label now has an explicit # XXX retranslation note beside the existing entry in the other language catalogs. Spanish was updated directly, and the new strings carry untranslated markers where appropriate. English remains the reference wording.

functions.inc.php
TrapoSAMA: Thanks for the suggestion and the privacy reminder. Implemented in 0f4f0bbe: red now means ‘Check local destination’, while blue means ‘External destination’. Both can appear for mixed target lists. Destination existence checks are restricted to the caller's authorized domains, including the target of an active alias domain; other administrators' address records are not inspected. Catch-all, recipient-delimiter, vacation and configured exceptions remain supported.

The wording deliberately avoids claiming guaranteed non-delivery because Postfix may have routing outside PostfixAdmin. Four focused PHP 8.4/SQLite tests passed (15 assertions), including unchanged output whether an inaccessible destination exists or not. The updated patch was also applied to local master and the authenticated Spanish legend checked. PR #1152 is being left unchanged until this master implementation has been reviewed.

TrapoSAMA:
The PostgreSQL failure in the first review run was caused by the new test fixture inserting aliases before creating their referenced domains. SQLite did not enforce that relationship, so the focused local run did not expose it.

Fixed in `28e54b1d`: the fixture now creates `local.test`, `alias.test`, and `private.test` before inserting aliases and removes the records in foreign-key-safe order during cleanup. The production logic is unchanged by this follow-up.

GitHub Actions run `34795496123` now passes lint, PHP 8.2–8.5, MySQL, and PostgreSQL. The focused local suite also remains green with 4 tests and 15 assertions.

TrapoSAMA:
Resolved the current master merge conflict in 8405ebc0. The only manual resolution was in CHANGELOG.TXT, preserving both the status-indicator entry and the newly merged maintenance entry. The pagination changes merged cleanly and retain the authorized-domain status call and the revised legend.

Focused validation passed: PHP lint for functions.inc.php, public/list-virtual.php, and model/AliasHandler.php; 4 AliasDestinationStatusTest tests with 15 assertions. PHPUnit reported only the expected missing coverage-driver warning. The refreshed production patch also applies cleanly to current master and reproduces GitHub's merge-ref tree exactly.

model/AliasHandler.php
```diff
@@ -436,7 +436,7 @@ protected function read_from_db_postprocess($db_result)
             }
 
             if ($this->struct['status']['display_in_list'] && Config::bool('show_status')) {
-                $db_result[$key]['status'] = gen_show_status($db_result[$key]['address']);
+                $db_result[$key]['status'] = gen_show_status($db_result[$key]['address'], is_array($this->allowed_domains) ? $this->allowed_domains : []);
```
cboltz: For some reason I don't remember, PFAHandler.php initializes (since 2011, see 68c79b90136f4e1177256a41d13d594e47dbcf44)

    protected $allowed_domains = false;

Would it make sense to change that to

    protected $allowed_domains = [];

to make the is_array check here superfluous?

config.inc.php
```diff
@@ -610,11 +610,12 @@ function x_mailbox_postcreation_action($id, $values) {
 // 'show_status_text' will be displayed with the background colors
 // associated with each status, you can customize it here
 $CONF['show_status_text']='&nbsp;&nbsp;';
-// show_undeliverable is useful if most accounts are delivered to this
-// postfix system.  If many aliases and mailboxes are forwarded
-// elsewhere, you will probably want to disable this.
+// Distinguish unrecognized destinations in managed domains from external targets.
```
cboltz: This comment might make sense in the context of this PR/change, but for average users it's confusing.

Maybe use something like

    Highlight undeliverable destinations in locally managed domains

(Line +614 is fine.)

functions.inc.php
```diff
@@ -2382,9 +2382,10 @@ function check_db_version($error_out = true)
  *        addresses list in show_custom_domains
  *
  * @param string $show_alias
+ * @param array $allowed_domains Domains authorized for the caller; empty means no destination lookups.
  * @return string
  */
-function gen_show_status($show_alias)
+function gen_show_status($show_alias, array $allowed_domains = [])
```
cboltz: Making the parameter optional means all targets will be marked as "external" (if $allowed_domains is missing), which is a confusing result.

Better ensure all callers hand over $allowed_domains, and make it a required parameter.

Also, in the parameter description, I'd add something like "Domains not listed in $allowed_domains will be marked as external."

functions.inc.php
cboltz: > The wording deliberately avoids claiming guaranteed non-delivery because Postfix may have routing outside PostfixAdmin.

While this is in theory possible, I'd call it unlikely - and would make the wording stronger. IMHO we can assume that PostfixAdmin-managed domains are purely managed in PostfixAdmin, without any special magic on the Postfix side.

If someone adds special magic on the Postfix side, this behaviour will be more confusing than wrongly claiming "undeliverable" ;-)

functions.inc.php
TrapoSAMA: Agreed. In 11c2ef23, allowed_domains is now required, and its parameter description states that domains not listed there are marked as external. Both production callers already pass their authorized domain arrays; the focused 4-test/15-assertion suite remains green.

model/AliasHandler.php
TrapoSAMA: Thanks. I kept the historical PFAHandler default unchanged to avoid widening this focused change across handlers. AliasHandler declares a domain field, so its constructor always populates allowed_domains with an array; 11c2ef23 therefore removes the defensive is_array() check and passes the property directly.

config.inc.php
TrapoSAMA: Updated in 11c2ef23 to "Highlight undeliverable destinations in locally managed domains." The following comment still clarifies that this checks PostfixAdmin records rather than actual delivery.

functions.inc.php
TrapoSAMA: Agreed. 11c2ef23 strengthens the base label to "Undeliverable local destination" and updates Spanish to "Destino local no entregable"; the other catalogs have refreshed retranslation notes. The short tooltip remains factual and explains that the address is not listed locally.

PR #1151, opened 2026-09-13
Show usage and neutral boxes for unlimited quotas

TrapoSAMA:
Unlimited mailboxes currently show only the unlimited label, hiding available usage. Unlimited domain counters also lose their quota box, and the domain storage label continues to show assigned quota even when actual usage is available.

Display available mailbox usage as `used / ∞` inside the existing neutral quota box. Keep unlimited domain alias and mailbox counts inside the same boxes, and display actual domain storage usage when quota tracking is enabled, retaining assigned quota in the tooltip. Limited and disabled quotas keep their existing behavior. Missing mailbox usage remains the translated unknown value; tracking disabled does not invent a usage value.

This ports the locally validated hotfix without unrelated production patches. It changes two templates, adds a changelog entry, and adds focused template regression coverage. No configuration, schema, dependency, authorization, or data-query changes are required. This PR targets master only.

Validation:
- PHP 8.4.23: 4 focused template tests, 52 assertions passed, covering unlimited, limited, disabled, zero/unknown usage, tracking disabled, and domain tooltips.
- PHP syntax, focused PHP-CS-Fixer dry-run, and git diff whitespace checks passed.
- The production hotfix applied cleanly with GNU patch; authenticated Spanish pages on local master returned HTTP 200 without template errors.
- The owner accepted the hotfix and requested this PR after local delivery.
- Full test suites, additional PHP/database matrices, and automated mobile/theme visual checks were not run locally; there are no SQL or migration changes.


templates/list-virtual_mailbox.tpl
```diff
@@ -70,7 +70,9 @@
                 {if $CONF.quota===YES}
                     <td>
                         {if $item.quota==0}
-                            {$PALANG.pOverview_unlimited}
+                            <div class="quota_bar">
+                                <span class="quota_label">{if $boolconf_used_quotas}{$divide_quota.current[$i]} / {/if}&infin;</span>
```
DavidGoodwin: `{else}&infin;{/if}` ?

TrapoSAMA: Thanks for checking this. I made both branches explicit: with usage tracking enabled the label remains `used / ∞`; otherwise it displays only `∞`. Moving infinity only into the else branch would leave the tracked-usage label ending in a slash.

Published in 96696b23, which also merges current master and resolves the changelog conflict while preserving both entries. The focused template checks pass on PHP 8.4: 4 tests, 52 assertions, including tracking enabled and disabled.

